PR #60, opened 2015-03-10
WIP:FS recon and mne BEM in python

kambysese:
@Eric89GXL this is a remedial attempt at pythonizing the FS reconstruction and mne BEM routines. Have yet to test and debug but thought you might have some comments, wdyt?


Eric89GXL:
@kambysese mne-python recently added a Python wrapper for `mne_flash_bem`, it  would be great if we could just extend that to work for our data somehow. And maybe the `freesurfer` commands should also go there, since most people need that step, too. WDYT?

kambysese:
@Eric89GXL I know this has been festering, and I also know that mne-python has implemented `mne_watershed_bem` which offers some options that our wrapper does not, and others that our wrapper has. I was considering integrating our wrapper into mne-python as a PR, but I am pretty reluctant about coding for that project. Otherwise I was planning to call `mne_watershed_bem` in the mnefun wrapper to clean things up a bit. As it stands our wrapper already calls `mne_flash_bem` to create 3-layers model.

Eric89GXL:
Fair enough, might be worth changing that to use `mne flash_bem` instead at least, so we stay in python-land as much as possible.

kambysese:
I think we're saying the same thing. There is an `mne_flash_bem_model` that I already use in our wrapper, and I think more recently an `mne_watershed_bem` that I can also use to clean up the code a bit around the `freesurfer` commands that people use here. Are we on the same page?

Eric89GXL:
No, I mean `mne flash_bem` (note the space) which means it uses mne-python's wrapper, instead of MNE-C's

kambysese:
I am using
```python
run_subprocess(['mne', 'flash_bem_model', '-s', subject, '-d', subjects_dir,
                        '--flash05', flash05, '--flash30', flash30, '-v'], env=this_env)
```
Doesn't `mne flash_bem` run the file mne/commnds/mne_flash_bem_model.py?

Eric89GXL:
Oh, yeah it might. Never mind then, that should be fine. Sorry for the confusion.

bin/run_mneBEM.py
```diff
@@ -0,0 +1,102 @@
...
+                           'for this script')
+
+    subj_path = op.join(subjects_dir, subject)
+    if not op.exists(subj_path):
+        raise RuntimeError('%s does not exits. Please check your subject '
+                           'directory path.' % subj_path)
+    
+    logger.info('1. Setting up MRI files and running watershed algorithm...')
+    run_subprocess(['mne_setup_mri', '--mri', 'T1', '--subject', subject, '--overwrite'], env=this_env)
+    run_subprocess(['mne_watershed_bem', '--subject', subject, '--overwrite'], env=this_env)
```
kambysese: @Eric89GXL  `make_make_morph_maps --to subject --from subject` what is that?

Eric89GXL: `mne_make_morph_maps` is an mne-C command that writes some files that make life easier

kambysese: Yes but `make_make_morph_maps --to subject --from subject` is a typo for something above and beyond `mne_make_morph_maps --to fsaverage --from subject`. Sorry I need to finish up this PR and am trying to address your comments from back in May.

Eric89GXL: oh sorry, I meant `mne_make_morph_maps --to subject --from subject` (i.e., to the subject itself)

PR #80, opened 2015-07-30
fix:import error with current mne-python upstream/master

kambysese:
@Eric89GXL not sure if this needs to be replaced or just deleted but no `_prep_channels` found in `mne.forward._make_forward` as of today's upstream.


Eric89GXL:
This will break the simulator, so for now I don't want to merge this. I plan on merging something soon that will fix it, yet allow the simulator to still work properly.

Eric89GXL:
Fixed by recent PR @kambysese

PR #81, opened 2015-07-30
Update analysis_fun.py

kambysese:
correct location for example raw data


Eric89GXL:
Thanks for the update

PR #83, opened 2015-08-06
FIX: Fix for newer mne-python and event len

Eric89GXL:
@staulu see if this fixes the event issue for you.


Eric89GXL:
@staulu I'm going to go ahead and merge since this fixes things at my end, will fix an import error for @kambysese, and you won't have time to look for a bit in the meantime. So once this is merged, you should be able to update your local `master` to get the changes, instead of checking out my branch. Feel free to re-open the issue that is about to auto-close if the fixes don't actually work for you.

PR #86, opened 2015-08-17
Fix bad

kambysese:
@Eric89GXL please review. I need this fix for auto bad channel detection in infant/kid cases where cHPI fails and MF sets channels to zero leading to spuriously high number of bad channels.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1368,7 +1368,7 @@ def do_preprocessing_combined(p, subjects):
             meg, eeg = 'meg' in raw, 'eeg' in raw
             picks = pick_types(raw.info, meg=meg, eeg=eeg, eog=False,
                                exclude=[])
-            flat = p.auto_bad_flat if isinstance(p.auto_bad_flat, dict)
+            flat = p.auto_bad_flat if isinstance(p.auto_bad_flat, dict) else None
             epochs = Epochs(raw, events, None, p.tmin, p.tmax,
                             baseline=_get_baseline(p), picks=picks,
                             reject=p.auto_bad_reject, flat=flat,
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1353,8 +1353,7 @@ def do_preprocessing_combined(p, subjects):
                 raise NameError('File not found (' + r + ')')
 
         bad_file = op.join(bad_dir, 'bad_ch_' + subj + p.bad_tag)
-        if isinstance(p.auto_bad, float):
-            assert type(p.auto_bad_reject) == dict
+        assert isinstance(p.auto_bad_reject, dict)
             print('    Creating bad channel file, marking bad channels:\n'
                   '        %s' % bad_file)
             if not op.isdir(bad_dir):
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1378,7 +1378,7 @@ def do_preprocessing_combined(p, subjects):
             if p.auto_bad_reject is None and p.auto_bad_flat is None:
                 raise RuntimeError('Auto bad channel detection active. Noisy and flat channel detection'
                                    'parameters not defined. '
-                                   'At least one criteria must be defined.')
+                                   'At least one criterion must be defined.')
             epochs = Epochs(raw, events, None, p.tmin, p.tmax,
                             baseline=_get_baseline(p), picks=picks,
                             reject=p.auto_bad_reject, flat=p.auto_bad_flat,
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1368,7 +1368,7 @@ def do_preprocessing_combined(p, subjects):
             meg, eeg = 'meg' in raw, 'eeg' in raw
             picks = pick_types(raw.info, meg=meg, eeg=eeg, eog=False,
                                exclude=[])
-            flat = p.auto_bad if type(p.auto_bad_flat) == dict else None
+            flat = p.auto_bad_flat if isinstance(p.auto_bad_flat, dict)
             epochs = Epochs(raw, events, None, p.tmin, p.tmax,
                             baseline=_get_baseline(p), picks=picks,
                             reject=p.auto_bad_reject, flat=flat,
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -1354,6 +1354,7 @@ def do_preprocessing_combined(p, subjects):
 
         bad_file = op.join(bad_dir, 'bad_ch_' + subj + p.bad_tag)
         if isinstance(p.auto_bad, float):
+            assert type(p.auto_bad_reject) == dict
```
Eric89GXL: `assert isinstance(p.auto_bad_reject, dict)`

mnefun/_mnefun.py
```diff
@@ -1368,10 +1369,10 @@ def do_preprocessing_combined(p, subjects):
             meg, eeg = 'meg' in raw, 'eeg' in raw
             picks = pick_types(raw.info, meg=meg, eeg=eeg, eog=False,
                                exclude=[])
-            assert type(p.auto_bad_reject) and type(p.auto_bad_flat) == dict
+            flat = p.auto_bad if type(p.auto_bad_flat) == dict else None
```
Eric89GXL: I don't quite understand this. If `p.auto_bad_flat` is a dict, use `p.auto_bad`, otherwise use `None`?

kambysese: The following line does epoching using p.auto_bad and p.auto_bad reject to get to the `badchs` via channels scores from the `Epochs` drop log. I am trying to give users the option to use `p.auto_bad` with or without rejection of trials where channels are flat i.e., `p.auto_bad_flat` in case e.g. when MF sets channels to zero after cHPI failure.

Eric89GXL: Did you mean `flat = `p.auto_bad_flat if isinstance(p.auto_bad_flat, dict) else None`? I'm still not quite sure on the logic. Right now you are setting it to `flat = p.auto_bad` sometimes, which is either a float or `None`.

kambysese: Yes `flat =p.auto_bad_flat if isinstance(p.auto_bad_flat, dict)` will work. I forgot it is set to `None` by default. I don't see where it's set to `float`. 
BTW my syntax checker is flagging `flat =p.auto_bad_flat if isinstance(p.auto_bad_flat, dict)` with `else` expected, is that so?

Eric89GXL: right now you have `flat = p.auto_bad`, not `flat = p.auto_bad_flat`, that's where the `float` can come from

yes you need the `else` clause in there

kambysese:
@Eric89GXL this should be correct now.

mnefun/_mnefun.py
```diff
@@ -1353,7 +1353,7 @@ def do_preprocessing_combined(p, subjects):
                 raise NameError('File not found (' + r + ')')
 
         bad_file = op.join(bad_dir, 'bad_ch_' + subj + p.bad_tag)
-        if isinstance(p.auto_bad, float):
+        assert isinstance(p.auto_bad_reject, dict)
```
Eric89GXL: I think you still want the original conditional `if isinstance(p.auto_bad, float)`

mnefun/_mnefun.py
```diff
@@ -1368,10 +1368,10 @@ def do_preprocessing_combined(p, subjects):
             meg, eeg = 'meg' in raw, 'eeg' in raw
             picks = pick_types(raw.info, meg=meg, eeg=eeg, eog=False,
                                exclude=[])
-            assert type(p.auto_bad_reject) and type(p.auto_bad_flat) == dict
+            flat = p.auto_bad_flat if isinstance(p.auto_bad_flat, dict) else None
```
Eric89GXL: I'm actually not sure you need to make this change. Can't the user just set `p.auto_bad_flat = None`? It seems like that should work.

mnefun/_mnefun.py
```diff
@@ -1368,10 +1368,10 @@ def do_preprocessing_combined(p, subjects):
             meg, eeg = 'meg' in raw, 'eeg' in raw
             picks = pick_types(raw.info, meg=meg, eeg=eeg, eog=False,
                                exclude=[])
-            assert type(p.auto_bad_reject) and type(p.auto_bad_flat) == dict
```
Eric89GXL: ...in which case this line could just be:
```
assert p.auto_bad_flat is None or isinstance(p.auto_bad_flat, dict)
assert p.auto_bad_reject is None or isinstance(p.auto_bad_reject, dict)
```

kambysese: That would be the more likely fix to the problem.  Initially `assert type(p.auto_bad_reject) and type(p.auto_bad_flat) == dict` failed if I didn't define `auto_bad_flat`. I think your recommendation with two assertions thoroughly covers all possible use cases.

Eric89GXL: Cool. I wonder if we should also raise an error or issue a warning if they are both `None` but `p.auto_bad is not None`, though, because in that case you're asking for auto bad detection but passing reject and flat params that do nothing. WDYT?

kambysese: That would be the prudent thing to do.

Eric89GXL: most prudent but not most expedient, so up to you whether you want to add it

kambysese:
```python
if p.auto_bad_reject is None and p.auto_bad_flat is None:
     warnings.warn('Auto bad detection active. But both flat and noisy channel criteria'
                              'are undefined. Using epoching rejection values to detect noisy'
                              'channels only. Check results.')
     p.auto_bad_reject = p.reject
```
WDYT?

Eric89GXL:
I think it's better to raise an issue, it forces the user to be more explicit about what they're doing

kambysese:
You mean `RuntimeError` or just a `Warning`?

Eric89GXL:
`RuntimeError`, yeah I meant raise an error, sorry

kambysese:
Alright. So much for a quick fix @Eric89GXL

Eric89GXL:
Thanks for sticking it out

kambysese:
That's going to be my Github headline "sticking it out". Thanks for putting up with it.

PR #87, opened 2015-08-17
Anatomy

kambysese:
@Eric89GXL Picking up where I left of on issue #60. Please review.

### DIFF
--- a/bin/run_fs_recon.py
+++ b/bin/run_fs_recon.py
@@ -5,10 +5,10 @@
 #          simplified bsd-3 license
 
 # TODO(ktavabi@gmail.com): Document
-'''Runs FreeSurfer recon-all on RMS combined multi echo MPRAGE volume.
+"""Runs FreeSurfer recon-all on RMS combined multi echo MPRAGE volume.
 
  example usage: python run_recon-all --subject subject --raw-dir ${SUBJECTS_DIR}/PARREC --openmp 2
-'''
+"""
 from __future__ import print_function
 
 import sys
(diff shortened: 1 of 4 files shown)


bin/run_fs_recon.py
```diff
@@ -0,0 +1,131 @@
...
+                           'for this script')
+
+    if op.isdir(op.join(subjects_dir, subject)) and not force:
+        raise RuntimeError('%s FreeSurfer directory exists. '
+                           'Use command line option --force to overwrite '
+                           'previous reconstruction results.' % subject)
+    if force:
+        shutil.rmtree(op.join(subjects_dir, subject))
+
+    os.mkdir(op.join(subjects_dir, subject))
```
Eric89GXL: I don't think you need this line, the next line should create it

bin/run_mne_bem.py
```diff
@@ -0,0 +1,109 @@
...
+    run_subprocess(['mne', 'make_scalp_surfaces', '--overwrite', '--subject', subject])
+    if op.isfile(op.join(subjects_dir, subject, 'bem/%s-head.fif' % subject)):
+        os.rename(op.join(subjects_dir, subject, 'bem/%s-head.fif' % subject),
+                  op.join(subjects_dir, subject, 'bem/%s-head-sparse.fif' % subject))
+    os.symlink((op.join(subjects_dir, subject, 'bem/%s-head-dense.fif' % subject)),
+               (op.join(subjects_dir, subject, 'bem/%s-head.fif' % subject)))
+
+    # Create source space
+    run_subprocess(['mne_setup_source_space', '--subject', subject, '--spacing', '%.0f' % 5, '--cps'],
+                   env=this_env)
```
Eric89GXL: I don't think we actually use this source space

Eric89GXL:
Other than my comments this looks reasonable to me. I assume you've been using it and it works?

bin/run_mne_bem.py
kambysese: What source space would you recommend? I've been using a 5mm spaced source space grid for my MEEG study and my fsaverage src space is a 5p7...

kambysese:
Yes it has worked for me with both three and single layer BEMs across two different datasets. I've not tested it on the example ds recently.

Eric89GXL:
It's alright, it's more important that it works on real data anyway. We can iterate in `master` if we need to.

Eric89GXL:
Thanks for the enhancement @kambysese

PR #88, opened 2015-08-18
ENH: Allow run indices to be provided

Eric89GXL:
@jyeatman see if this solves your use case. You can now do:
```
p.subject_run_indices = [None, [0, 1, 3, 4]]
```
This would use all runs for subject 0, and runs 0, 1, 3, and 4 for subject 1.

I am currently testing that the example code still works, but hopefully it will work for you already.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1356,7 +1356,8 @@ def _raw_LRFCP(raw_names, sfreq, l_freq, h_freq, n_jobs, n_jobs_resample,
         r = Raw(rn, preload=True)
         r.load_bad_channels(bad_file, force=force_bads)
         if sfreq is not None:
-            r.resample(sfreq, n_jobs=n_jobs_resample)
+            with warnings.catch_warnings(record=True):  # resamp of stim ch
+                r.resample(sfreq, n_jobs=n_jobs_resample)
         if l_freq is not None or h_freq is not None:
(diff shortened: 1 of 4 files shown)


mnefun/_mnefun.py
```diff
@@ -364,6 +367,11 @@ def do_processing(p, fetch_raw=False, do_score=False, push_raw=False,
              'Status',
              ]
     score_fun = p.score if p.score is not None else default_score
+    if len(inspect.getargspec(score_fun).args) == 2:
+        score_fun_two = score_fun
+
+        def score_fun(p, subjects, run_indices):
```
kambysese: @Eric89GXL does this also address the issue during scoring or is that on the user to resolve? I can see `run_indices` passed in but don't see it used. I am getting an I/O error during scoring for a subject with missing runs.

Eric89GXL: If you're using the built-in simple scoring, it works. If you wrote your own scoring script, you need to modify it to take in the `run_indices` as a third argument, and deal with it yourself. But the built-in scoring functions shows how it can be done.

kambysese: @jyeatman can you handle this and test with nlr subjects?

kambysese:
+1 for merge

Eric89GXL:
Thanks @kambysese

PR #89, opened 2015-08-19
FIX:close psd plotting window to prevent memory consumption

kambysese:
@Eric89GXL I got a memory usage warning about too many open `pyplot` windows during psd plotting for a large group of subjects. This is the fastest way I know to make it go away. LMKWYT.


Eric89GXL:
That should work. FYI it's better to use the non-global functions, so doing `fig.close()` or whatever the equivalent would be is slightly better. But this should be fine.

PR #91, opened 2015-08-27
FIX:status reporting

kambysese:
@Eric89GXL caught this bug trying to `print_status` as a function of `run_indices`; instead of using individual subject's runs list the entire group list of list was being passed to `get_raw_fnames` causing empty lists of files to be returned. This fixes the case for a list of runs list and works with example script.

### DIFF
--- a/mnefun/_status.py
+++ b/mnefun/_status.py
@@ -25,13 +25,13 @@ def print_proc_status(p, subjects, structurals, analyses, run_indices):
     # XXX TODO in subsequent PR:
     # * Add modified-date checks to make sure provenance is correct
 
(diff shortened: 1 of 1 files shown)


Eric89GXL:
Whoops! Yep this looks better.

kambysese:
Thanks.

------------------------------------------------------------
Kambiz Tavabi PhD
Institute for Learning & Brain Sciences
1715 Columbia Road N
Portage Bay Building
Box 357988
University of Washington
Seattle, WA 98195-7988
Tel: 206-685-6173
------------------------------------------------------------

PR #92, opened 2015-08-29
FIX:raw plotting

kambysese:
@Eric89GXL this tries to improve handling of ExG event files based on their availability during plotting raw data subsequent to preprocessing. It worked without error on my data with just EOG events as well as example data. However for some reason the plot is not being shown for some reason in my IDE, I am assuming some resurfacing pyplot and/or matplotlib backend issue. Can you run with example analysis with `params.plot_raw=True`?

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1792,26 +1792,19 @@ def _viz_raw_ssp_events(p, subj, ridx):
                 if ri in p.get_projs_from]
     all_proj = op.join(pca_dir, 'preproc_all-proj.fif')
     projs = read_proj(all_proj)
-    ev = []
-    for n in ['ecg', 'blink']:
+    colors = dict()
+    ev = np.zeros((0, 3), int)
+    for n, c, cid in zip(['ecg', 'blink'], ['r', 'b'], [999, 998]):
         fname = op.join(pca_dir, 'preproc_%s-eve.fif' % n)
         if op.isfile(fname):
-            ev.append(read_events(fname))
-    if len(ev) == 2:
-        ev = np.concatenate((ev[0], ev[1]))
-        ev = ev[np.argsort(ev[:, 0], axis=0)]
-        colors = {int(idd): c for c, idd in zip(['r', 'b'], np.unique(ev[:, 2]))}
-    else:
-        ev = ev[0]
-        colors = {int(np.unique(ev[:, 2])[0]): 'r'}
+            ev = np.concatenate((ev, read_events(fname)))
+            colors[cid] = c
+    ev = ev[np.argsort(ev[:, 0], axis=0)]
     raw = _raw_LRFCP(pre_list, p.proj_sfreq, None, None, p.n_jobs_fir,
                      p.n_jobs_resample, projs, None, p.disp_files,
                      method='fft', filter_length=p.filter_length,
                      force_bads=False)
-    if len(colors) == 0:
-        raw.plot(show=True)
-    else:
-        raw.plot(events=ev, event_color=colors, show=True)
+    raw.plot(events=ev, event_color=colors)
 
 
 def gen_html_report(p, subjects, structurals, run_indices=None,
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -1792,19 +1792,20 @@ def _viz_raw_ssp_events(p, subj, ridx):
                 if ri in p.get_projs_from]
     all_proj = op.join(pca_dir, 'preproc_all-proj.fif')
     projs = read_proj(all_proj)
-    ev_names = [op.join(pca_dir + '/' + ii) for ii in
-                ['preproc_ecg-eve.fif', 'preproc_blink-eve.fif']]
+    ev_names = glob.glob(op.join(pca_dir, 'preproc*-eve.fif'))
```
Eric89GXL: we can probably leave the old check, it's a bit more explicit than globbing. Users for example might reasonable rename old versions in some way that the `glob` would catch.

Then  later, instead of doing the `assert len(ev) == 2`, you can do two separate `if op.isifle(...)` checks.

Does that sound alright?

mnefun/_mnefun.py
```diff
@@ -1792,19 +1792,20 @@ def _viz_raw_ssp_events(p, subj, ridx):
...
+    else:
+        ev = ev[0]
+        colors = {int(np.unique(ev[:, 2])[0]): 'r'}
     raw = _raw_LRFCP(pre_list, p.proj_sfreq, None, None, p.n_jobs_fir,
                      p.n_jobs_resample, projs, None, p.disp_files,
                      method='fft', filter_length=p.filter_length,
                      force_bads=False)
-    raw.plot(events=ev, event_color={999: 'r', 998: 'b'})
-    plt.draw()
-    plt.show()
```
Eric89GXL: Does the plotting in your IDE work if you put this `plt.draw()` or `plt.show()` back in?

kambysese: Nope.

mnefun/_mnefun.py
kambysese: yep.

kambysese:
Newer fix explicitly checks for ExG event files before determining events and colors for plotting raw and handles case where there are no events to plot at all. LMKWYT.

mnefun/_mnefun.py
```diff
@@ -1792,19 +1792,26 @@ def _viz_raw_ssp_events(p, subj, ridx):
...
+        fname = op.join(pca_dir, 'preproc_%s-eve.fif' % n)
+        if op.isfile(fname):
+            ev.append(read_events(fname))
+    if len(ev) == 2:
+        ev = np.concatenate((ev[0], ev[1]))
+        ev = ev[np.argsort(ev[:, 0], axis=0)]
+        colors = {int(idd): c for c, idd in zip(['r', 'b'], np.unique(ev[:, 2]))}
+    else:
+        ev = ev[0]
+        colors = {int(np.unique(ev[:, 2])[0]): 'r'}
```
Eric89GXL: sine you already have the loop, it's better just to do all the processing there. For example:
```
colors = dict()
ev = np.zeros((0, 3), int)
for n in [...]:
    if op.isfile(...)
        ev = np.concatenate((ev, ...))
        colors[...] = ...
ev = ev[np.argsort(ev[:, 0], axis=0)]
```
No extra conditionals would be needed in this case. You also *shouldn't* need the `if len(colors) == 0` check later, either, since `mne-python` should properly handle zero-length `events` (if it doesn't it should be fixed at the `mne-python` end)

Eric89GXL:
Looks good now, I assume you've tried it and it fixes your use case?

Eric89GXL:
(other than the IDE plotting badness, which I think is a separate issue)

kambysese:
yes it fixes my use case.

kambysese:
BTW do you have any idea how to track a commit?

Eric89GXL:
What do you mean?

kambysese:
Alex is asking whether I can track the commit that may be associated with my IDE troubles with plotting; I have no idea what he is asking for.

drammock:
Probably he means checking out progressively older versions until you find
one where the problem goes away, then working forward until you find where
it breaks. The most common approach is "divide and conquer" also known as
"git bisect".

Eric89GXL:
Yeah, that's likely what he means. Or the anaconda library versions, in this case probably `matlplotlib`. I'll reply more in that issue on `mne-python`. Thanks for these fixes @kambysese

PR #93, opened 2015-09-02
FIX:plot window process blocking behavior.

kambysese:
@Eric89GXL This works for me, and I am completely happy with that if you feel like merging :)


mnefun/_mnefun.py
```diff
@@ -1804,7 +1804,7 @@ def _viz_raw_ssp_events(p, subj, ridx):
                      p.n_jobs_resample, projs, None, p.disp_files,
                      method='fft', filter_length=p.filter_length,
                      force_bads=False)
-    raw.plot(events=ev, event_color=colors)
+    raw.plot(events=ev, event_color=colors, block=True)
```
Eric89GXL: So we do want blocking for some functions, and not others, then? Can you add a `block=False` for the one @jyeatman is having trouble with?

PR #100, opened 2015-11-03
fixed import error

mdclarke:
@Eric89GXL please review encountered import error on mne-python after updating both repos to current master.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -3,7 +3,6 @@
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 
 from __future__ import print_function
-
 import os
 import os.path as op
 import inspect
@@ -13,25 +12,28 @@ import subprocess
 import glob
 from collections import Counter
 from time import time
-
 import numpy as np
 from scipy import io as spio
 import matplotlib.pyplot as plt
 from numpy.testing import assert_allclose
-
 from mne import (compute_proj_raw, make_fixed_length_events, Epochs,
                  find_events, read_events, write_events, concatenate_events,
                  read_cov, write_cov, read_forward_solution,
-                 compute_covariance,
                  write_proj, read_proj, setup_source_space,
                  make_forward_solution, get_config, write_evokeds,
                  make_sphere_model, setup_volume_source_space,
                  read_bem_solution)
+
+try:
+    from mne import compute_raw_covariance  # up-to-date mne-python
+except ImportError:  # oldmne-python
+    from mne import compute_raw_data_covariance as compute_raw_covariance
 from mne.preprocessing.ssp import compute_proj_ecg, compute_proj_eog
 from mne.preprocessing.maxfilter import fit_sphere_to_headshape
 from mne.minimum_norm import make_inverse_operator
 from mne.label import read_label
 from mne.epochs import combine_event_ids
+
 try:
     from mne.chpi import _quat_to_rot, _rot_to_quat
 except ImportError:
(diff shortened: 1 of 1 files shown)


Eric89GXL:
This will work for people who have updated, but it's best not to force people to update if they don't have to. We also need to update the calls to `compute_raw_data_covariance` to refer to the new function. How about something like this?
```
try:
    from mne import compute_raw_covariance  # up-to-date mne-python
except ImportError:  # oldmne-python
    from mne import compute_raw_data_covariance as compute_raw_covariance
```
And then look for the instance(s) of `compute_raw_data_covariance` and replace them with `compute_raw_covariance`.

kambysese:
@Eric89GXL is there a particular reason why this PR is outstanding?

Eric89GXL:
Yeah, I didn't know there was a follow up commit :) @mdclarke one of the (too) many github quirks is that commits are not notified, so it's standard practice to request re-review once comments are addressed. I should be able to look in a bit.

mnefun/_mnefun.py
```diff
@@ -3,7 +3,6 @@
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 
 from __future__ import print_function
-
```
Eric89GXL: this was actually intentional to separate `future` imports from standard imports

mnefun/_mnefun.py
```diff
@@ -13,25 +12,28 @@
 import glob
 from collections import Counter
 from time import time
-
```
Eric89GXL: ...and standard library imports from other-library imports

mnefun/_mnefun.py
```diff
@@ -13,25 +12,28 @@
...
 import numpy as np
 from scipy import io as spio
 import matplotlib.pyplot as plt
 from numpy.testing import assert_allclose
-
```
Eric89GXL: ...and other library imports from `mne` imports

mnefun/_mnefun.py
```diff
@@ -45,7 +47,6 @@
 from mne.utils import run_subprocess
 from mne.report import Report
 from mne.io.constants import FIFF
-
```
Eric89GXL: ...and `mne` imports from `mnefun` (local) imports

mnefun/_mnefun.py
```diff
@@ -936,7 +938,7 @@ def fix_eeg_files(p, subjects, structurals=None, dates=None, run_indices=None):
         names = [name for name in raw_names if op.isfile(name)]
         # noinspection PyPep8
         if structurals is not None and structurals[si] is not None and \
-                dates is not None:
+                        dates is not None:
```
Eric89GXL: Did you editor complain about this line? The original should be okay according to PEP8, whereas this should be a violation (not a multiple of 4 indent)

Eric89GXL:
Sorry to nitpick this simple change to death, but it's good to standardize on style issues like this from the get-go. Otherwise LGTM.

mdclarke:
@Eric89GXL Sorry! just started using PyCharm, might have mashed some buttons.. also Kam drinks when he codes.

Eric89GXL:
Looks good now except for the unaddressed comments above about removed newlines.

kambysese:
WTH?! I am sure we put those in. I ran a pep8 inspection on the indentation issue and the fix must have reformated the imports. @mdclarke pycharm editor still needs some tweaks. Sorry about that @Eric89GXL. Will push again tomorrow.

mdclarke:
@Eric89GXL ready for merge.

Eric89GXL:
Thanks @mdclarke

PR #102, opened 2015-11-09
Check rotation matrix is special orthogonal

kambysese:
@Eric89GXL I think this is working now, will continue to test with infant data. @mdclarke what was the bad_baby subject giving you an error for median hp calc?


Eric89GXL:
Thanks @kambysese

PR #103, opened 2015-11-23
Update _mnefun.py

kambysese:
@Eric89GXL fixed missing `compute_covariance` module import. Thinking this was deleted during last PR concerning raw data covariance.


Eric89GXL:
Whoops!

PR #104, opened 2015-12-17
Anatomy

kambysese:
@Eric89GXL LMK if you're okay with this PR or if I should split it up. It contains some minor fixes to my wip anatom/bem shells that another user is now testing, along with a minor fix to fix a TypeError I encountered with call to `_empty_info` during median hp calculation for split up raw files. at line 601 of _mnefun.py.


bin/run_fs_recon.py
```diff
@@ -122,10 +121,11 @@ def _run(subjects_dir, subject, raw_dir, force, mp, volume):
     run_subprocess(['mri_concat', '--rms', '--i', mri,
                     '--o', op.join(subjects_dir, subject, 'mri/orig/001.mgz')],
                    env=this_env)
-    run_subprocess(['recon-all', '-openmp', mp, '-subject', subject, '-all'], env=this_env)
+    run_subprocess(['recon-all', '-openmp', '%.0f' % mp, '-subject', subject, '-all'], env=this_env)
```
Eric89GXL: FYI if you want to force an integer you can do `'%d' % mp`. But since you know it's an integer you can also just do `str(mp)`. But this should work too

Eric89GXL:
Thanks @kambysese

bin/run_fs_recon.py
kambysese: Thanks @Eric89GXL I actually tried several variants to get this to work, and this was the one that `run_subprocess` appeared to like. Will keep it in mind and let you know if I encounter `TypeErrors` again.

Eric89GXL: I have a feeling you tried passing `mp` itself, not cast to a `str` already. `run_subprocess` wants them all to be strings. You might have also tried `%f` or something that would have made it a float and thus given the string `'1.0'` or something, which would fail at the `recon-all` stage

PR #107, opened 2016-01-21
WIP: SSS in python

kambysese:
@Eric89GXL added support for maxwell filtering with mne-python and tested with example data. Made a number of changes I'd like to hear if you're ok with or not. LMKWYT at your earliest convenience. cc @mdclarke

PR #112, opened 2016-02-03
Maxwell filtering

kambysese:


kambysese:
@Eric89GXL I finally managed to resolve the work in #107 with current master. If this new PR is okay with you? I will go through the list of comments and try to address concerns you raised.

Eric89GXL:
That commit history looks tortured. Yeah it's fine if you work off of this one instead :)

kambysese:
Lmao

kambysese:
@Eric89GXL please review. I will try and unify maxfilter and maxwell_filtering arguments in a fu pr.

mnefun/_mnefun.py
```diff
@@ -854,6 +895,83 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
...
+        #  process raw files
+        for ii, (r, o) in enumerate(zip(raw_files, raw_files_out)):
+            if not op.isfile(r):
+                raise NameError('File not found (' + r + ')')
+            raw = Raw(r, preload=True, allow_maxshield=True)
+            raw.load_bad_channels(prebad_file, force=True)
+            raw.fix_mag_coil_types()
+            raw = filter_chpi(raw)
+            _, origin, _ = fit_sphere_to_headshape(raw.info)
+            origin /= 1000
```
Eric89GXL: you shouldn't need to do this, this is equivalent to `origin='auto'` in `maxwell_filter`

Eric89GXL: (and doesn't this thus ignore the `p.sss_origin` argument?)

mnefun/_mnefun.py
```diff
@@ -854,6 +895,83 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
...
+            # apply maxwell filter
+            raw_sss = maxwell_filter(raw, origin=origin, int_order=p.int_order, ext_order=p.ext_order,
+                                     calibration=cal_file, cross_talk=ct_file,
+                                     st_correlation=p.st_correlation, st_duration=st_duration,
+                                     destination=trans_to, coord_frame='head',
+                                     head_pos=pos, regularize=reg)
+
+            raw_sss.save(o, overwrite=True, buffer_size_sec=None)
+            #  process erm files if any
+            if len(erm_files) > 1:
```
Eric89GXL: I think you mean > 0 not > 1, and you don't actually need the conditional because if `len(erm_files) == 0` then the `for` loop won't even execute

mnefun/_mnefun.py
```diff
@@ -1458,7 +1576,7 @@ def _raw_LRFCP(raw_names, sfreq, l_freq, h_freq, n_jobs, n_jobs_resample,
         print('    Loading and filtering %d files.' % len(raw_names))
     raw = list()
     for rn in raw_names:
-        r = Raw(rn, preload=True)
+        r = Raw(rn, preload=True, allow_maxshield=allow_maxshield)
```
Eric89GXL: you need to wrap this in `with warnings.catch_warnings(record=True):` to catch the warning

Eric89GXL:
Otherwise looks reasonable to me. Might be good to add a note in the Params docstring about which parameters are only used or valid in Python as a reminder about what to fix later.

kambysese:
@Eric89GXL issues addressed, will work on doc string. Modified `run_sss_positions' to call mf explicitly as it appears some users @jyeatman do not have an alias for it.

jyeatman:
I run into permissions issue. I suggest adding in a chmod 777 command to make the file on kasga easy to deal with

jyeatman:
@Eric89GXL @kambysese If you add something like
cmd = ['scp', '-P' + port] + fnames_in + [host + ':~/'] + [' && ssh' + host + '"chmod 777 ~/' + fnames_in + '"']
to each of the scp commands then it will deal with the permissions issues within the same command. I started editing the file but I realized that I don't know all the places this may cause issues

Eric89GXL:
Which file has problems? I suspect you need to be in the right group
(neuromag). Can you check to see if you are? 777 isn't the safest from a
security point of view.

jyeatman:
I suspect that there is more going on here, probably having to do with the
correspondence between user IDs. But in general, I am a proponent of all
tmp files being written with 777 permissions since these files are intended
to be deleted anyway.

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

Eric89GXL:
That would be a quick solution, so if you're pressed for time we can do it.
To me it's indicative of some likely underlying issue with what we're
doing, though. I suspect there is something to be learned here. I'd prefer
spending a bit of time trying to fix it, and use full permissions as a last
resort if possible.

What is the traceback of the error?

jyeatman:
That actually gets to a larger question - Isn't the whole scping thing a
temporary solution? Don't we want to strive for a full, open source, python
pipeline that doesn't require messy file transfers and system calls? Here's
the error

(u'',
u'/mnt/diskArray/projects/MEG/nlr/mnetest/mneanalysis/202_dd150827/sss_fif/202_dd150827_1_raw_sss.fif:
Permission denied\n')
  File "/home/jyeatman/git/BrainTools/projects/NLR_MEG/nlr_testmnefun.py",
line 120, in <module>
    gen_report=False
  File
"/home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg/mnefun/_mnefun.py",
line 555, in do_processing
    outs[ii] = func(p, subjects, run_indices)
  File
"/home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg/mnefun/_mnefun.py",
line 724, in run_sss
    run_sss_localy(p, subjects, run_indices)
  File
"/home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg/mnefun/_mnefun.py",
line 946, in run_sss_localy
    pos = _headpos(p, subj, r, o)
  File
"/home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg/mnefun/_mnefun.py",
line 2106, in _headpos
    run_sss_positions(p, file_in, file_out)
  File
"/home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg/mnefun/_mnefun.py",
line 890, in run_sss_positions
    run_subprocess(cmd)
  File "<string>", line 2, in run_subprocess
  File "/home/jyeatman/git/mne-python/mne/utils.py", line 551, in verbose
    return function(*args, **kwargs)
  File "/home/jyeatman/git/mne-python/mne/utils.py", line 869, in
run_subprocess
    raise subprocess.CalledProcessError(p.returncode, command, output)
subprocess.CalledProcessError: Command '['scp', '-P22',
'jason@kasga.ilabs.uw.edu:~/temp_1454972911.2_hp.txt',
'/mnt/diskArray/projects/MEG/nlr/mnetest/mneanalysis/202_dd150827/sss_fif/202_dd150827_1_raw_sss.fif']'
returned non-zero exit status 1


Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

Eric89GXL:
> Isn't the whole scping thing a temporary solution? Don't we want to strive for a full, open source, python pipeline that doesn't require messy file transfers and system calls?

Yeah, but there isn't really a known timeline for getting the position estimation working as well in Python as it does in MaxFilter. It's not a high priority currently.

Unfortunately that error trace doesn't actually say what the error was, permission-related or otherwise. What makes us think it's a permissions issue, as opposed to e.g. the file not existing? Does the file exist on `kasga`?

jyeatman:
Yeah the file exists and has weird permissions

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

Eric89GXL:
I'd look but I can't access your home dir -- what are the permissions? Even
just having `r` for user should really be enough here, since it's logging
in as your user. The file is produced by MaxFilter so I wonder if they have
some bug with their file creation. You are in the "neuro" group, so that
shouldn't be the problem.

Kam, does this function work for you?

kambysese:
My guess is that some users are inadvertently set up differently on KASGA, e.g. "neuro" group membership. I know that Maggie encounters restricted permissions on the run_sss.sh during pushes, which I believeJason has also seen, correct? It might be worth talking to the sys admin...

jyeatman:
I have that same issue with run_sss.sh So I always have to ssh in
afterwards and change the permissions and then re-run

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

Eric89GXL:
@jyeatman are you on OSX or Linux? It's possible `scp` is somehow copying with weird permissions. What are the permissions on `~/temp_1454972911.2_hp.txt`?

jyeatman:
I'm Linux. And my guess is that the weird permissions come from the fact
that I have a different user ID on kasga versus my local machine so the scp
command doesn't know what to do in terms of file ownership. Alternatively I
could modify my directory permissions. But it still might be easiest to
explicitly set permissions for each transfer since different users or
administrators might arbitrarily set up different permissions.

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

Eric89GXL:
It should just take the user ID of the user you're loggin in as...

What are the permissions on ~/temp_1454972911.2_hp.txt?

jyeatman:
yep that does happen

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

jyeatman:
Okay it has to do with differences in groups and usernames across the
systems. When the files get scped back to my local machine, they end up
with a random group ownership that has weird permissions set. My guess is
that this is because the group ID number on kasga corresponds to some
random group ID on my local machine. Is there any foolproof way of denoting
the proper permissions and groupIDs to be assigned to files in the scp call?

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

Eric89GXL:
Can you report the actual permissions and user/group of the file, e.g. by `ls -l`? It would help diagnose the problem.

Also see e.g.:

http://unix.stackexchange.com/questions/144596/scp-and-file-owner

One possible fix would be to `chown user:user` on the file before transferring to the local machine, but hopefully it's not necessary. According to the link above, the file should be given proper permissions at the receiving end. I guess that means there's some problem at the server end, but trying to figure it out without seeing the actual permission line is tough.

jyeatman:
Okay, it looks like setting my home directory permissions on kasga
chmod -R 750 /home/jason
worked.

But now I get
IOError: File
"/mnt/diskArray/projects/MEG/nlr/mnetest/mneanalysis/202_dd150827/raw_fif/202_dd150827_raw.pos"
does not exist

Shouldn't this be scp'd back over?

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
@Eric89GXL 'run_sss_position'  is not working as expected (1) it writes out quaternion text file as .fif file in (2) raw_fif dir.  Don't we want a .pos or .txt file in sss_log dir?

Eric89GXL:
> Shouldn't this be scp'd back over?

Yeah, it should be copied over somewhere. If it's not, there must be some bug with this PR or the position calculation function. @kambysese can you investigate? It seems related to the points you brought up.

> it writes out quaternion text file as .fif file

You mean `-quat.fif` file? That's not a position file, it's a copy of the raw file with the quaternion channels added. It shouldn't actually copy that over. I don't think it did for me when I ran the function. It copies it to the raw dir for you?

> in raw_fif dir

This is where the `.pos` files always used to live. The positions are more or less a property of the raw data, so it's a reasonable place to be. And eventually they will be produced by `mne-python`'s CHPI fitting routines instead of MaxFilter i.e. not an SSS process really.

kambysese:
This "should" work now, and "should" be ready for merge. Seriously, it works on my end, please review.

kambysese:
@Eric89GXL prior to most recent commit the hp text file from maxfilter was being written to the `raw_fif` dir but with the same name as the raw input file.

mnefun/_mnefun.py
```diff
@@ -801,19 +844,19 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
             pass
 
 
-def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
+def run_sss_positions(p, fname_in, fname_out, opts=''):
```
Eric89GXL: Don't change the signature of these `run_` functions. They're designed to be used outside of `analysis` scripts, as standalone calls.

mnefun/_mnefun.py
```diff
@@ -854,6 +898,83 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
     run_subprocess(cmd)
 
 
+def run_sss_localy(p, subjects, run_indices):
```
Eric89GXL: this one is okay to rely on `params`, though, because it's only meant to be used with the analysis scripts

mnefun/_mnefun.py
```diff
@@ -952,7 +1073,7 @@ def fix_eeg_files(p, subjects, structurals=None, dates=None, run_indices=None):
         names = [name for name in raw_names if op.isfile(name)]
         # noinspection PyPep8
         if structurals is not None and structurals[si] is not None and \
-                dates is not None:
+                        dates is not None:
```
Eric89GXL: ?

mnefun/_mnefun.py
```diff
@@ -1775,7 +1897,7 @@ def apply_preprocessing_combined(p, subjects, run_indices):
             _viz_raw_ssp_events(p, subj, run_indices[si])
 
 
-class FakeEpochs():
+class FakeEpochs:
```
Eric89GXL: This should probably be `FakeEpochs(object)` to be correct

mnefun/_mnefun.py
```diff
@@ -400,6 +436,9 @@ def do_processing(p, fetch_raw=False, do_score=False, push_raw=False,
         Print status (determined from file structure).
     """
     # Generate requested things
+    if p.sss_type is 'python':
```
Eric89GXL: don't use `is` for string comparison use `==`

Eric89GXL:
Otherwise LGTM @kambysese.

> Shouldn't this be scp'd back over?

@jyeatman are you running on this branch and hitting this problem? Maybe Kam's latest commits fix it?

jyeatman:
I'll give it a try when I get into the office today

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

jyeatman:
Naming error.
IOError: File
"/mnt/diskArray/projects/MEG/nlr/mnetest/mneanalysis/202_dd150827/raw_fif/202_dd150827_raw.pos"
does not exist

The file is actually titled
/mnt/diskArray/projects/MEG/nlr/mnetest/mneanalysis/202_dd150827/raw_fif/202_dd150827_1_raw.pos

should be an easy fix


Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
@jyeatman what's the trace on this?

jyeatman:
Same spot as before.

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
What is the name of the original raw file 202_dd150827_raw.fif or 202_dd150827_1_raw.fif? I don't see how why the file base names are different.

kambysese:
@Eric89GXL in the case of raw-1.fif datasets I am assuming a single hp file will work and `io.Raw` will handle the concatenation before `maxwell_filter`, correct?

Eric89GXL:
I think it's going to be more annoying than that. I think you'll need to get two `pos` files, one from the `_raw.fif` and the other from the  `_raw-1.fif` file, because MaxFilter won't process them as a single file. Then you'll need to combine the `pos` files (probably in memory using `read_head_pos` a couple of times) before passing them as a `head_pos` parameter to `maxwell_filter`, because it *will* treat them as a single instance.

kambysese:
@Eric89GXL please review. Tested with example data and my own (with splits). Made a quite a bit of change to `run_sss_position` to handle single files as input. Split files are handled by `run_sss_locally` which should also now fix the last error @jyeatman brought up. These changes would be reduced to 1-2 lines if and when we use the hp estimation function in mne-python.

mnefun/_mnefun.py
```diff
@@ -681,35 +719,40 @@ def _check_trans_file(p):
...
-        print(s)
-        print('-' * len(s))
-        run_subprocess(cmd, stdout=None, stderr=None)
-        print('-' * 70, end='\n\n')
+def run_sss(p, subjects, run_indices):
+    """Run SSS preprocessing remotely (only designed for *nix platforms) or
+    locally using Maxwell filtering in mne-python"""
+    if p.sss_type is 'python':
+        print(' Applying SSS locally using mne-python')
+        run_sss_localy(p, subjects, run_indices)
```
Eric89GXL: locally

mnefun/_mnefun.py
```diff
@@ -801,7 +844,7 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
             pass
 
 
-def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
+def run_sss_positions(fname_in, fname_out, host, port=22, opts=''):
```
Eric89GXL: why delete the default and change arg order?

mnefun/_mnefun.py
```diff
@@ -812,8 +855,6 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
     fname_out : str
         Output filename to use to store the resulting head positions
         on the local machine.
-    host : str
-        The SSH/scp host to run the command on.
```
Eric89GXL: ?

mnefun/_mnefun.py
```diff
@@ -822,38 +863,141 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
         raise IOError('input file not found: %s' % fname_in)
     if not op.isdir(op.dirname(op.abspath(fname_out))):
         raise IOError('output directory for output file does not exist')
-    fnames_in = [fname_in]
-    for ii in range(1, 11):
-        next_name = op.splitext(fname_in)[0] + '-%s' % ii + '.fif'
-        if op.isfile(next_name):
-            fnames_in.append(next_name)
-        else:
-            break
```
Eric89GXL: Why did you have to delete this? Did the detection not work properly before?

Eric89GXL: It seems like getting the autodetection working properly would be a better solution if we can get it working...

mnefun/_mnefun.py
kambysese: ?

mnefun/_mnefun.py
kambysese: the default arg 'kasga' was not working.

mnefun/_mnefun.py
kambysese: I figured if `host = p.sws_sss` or the remote identified by params this argument doesn't need to be explicit.

mnefun/_mnefun.py
kambysese: This func was quite buggy as it was, e.g. this particular loop was unclear to and I suspect the source of the most recent error @jyeatman reported concerning missing raw files with modified basenames as per `next_name`. WDYM by auto detection? The `dict` implementation should account for split files > raw-1.fif for a given run_index. In the past, e.g., in `get_raw_names` we resorted to reg expression to find split files. This loop wasn't really working, and by the time I figured out what you were trying to do, I had the `dict` approach working.

mnefun/_mnefun.py
Eric89GXL: If you have your `routes` configured properly to work on UW's network it should. Please revert

mnefun/_mnefun.py
Eric89GXL: you have "localy" in the function name it should be "locally"

mnefun/_mnefun.py
Eric89GXL: The docstring is useful to have when using this function outside of analysis scripts, as I have been doing...

mnefun/_mnefun.py
Eric89GXL: @kambysese this function is used outside of the context of analysis scripts though. What it was supposed to do (before these changes) was take a given input file name, see if any `-1` files exist, and automatically *also* process those with MaxFilter to get the positions. Make sense now?

jyeatman: The new version of the code is working for me so far

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

mnefun/_mnefun.py
kambysese: WDYM by `routes`? Are you somehow aliasing SSH connections?

mnefun/_mnefun.py
kambysese: I am beginning to wonder if I began working on the same cp of the `run_sss_positions` code!

kambysese:
@Eric89GXL reverted to previous version. Works with e.g, data. HP data for split files not supported.
@jyeatman could you test again with your data.

Eric89GXL:
> HP data for split files not supported.

You mean the function `run_sss_positions` does not work properly?

jyeatman:
Using fine calibration sss_cal.dat
  File "/home/jyeatman/git/BrainTools/projects/NLR_MEG/nlr_testmnefun.py",
line 120, in <module>
    gen_report=False
  File
"/home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg/mnefun/_mnefun.py",
line 558, in do_processing
    outs[ii] = func(p, subjects, run_indices)
  File
"/home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg/mnefun/_mnefun.py",
line 727, in run_sss
    run_sss_localy(p, subjects, run_indices)
  File
"/home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg/mnefun/_mnefun.py",
line 982, in run_sss_localy
    head_pos=pos, regularize=reg)
  File "<string>", line 2, in maxwell_filter
  File "/home/jyeatman/git/mne-python/mne/utils.py", line 551, in verbose
    return function(*args, **kwargs)
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
269, in maxwell_filter
    head_frame, ignore_ref)
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
1576, in _update_sensor_geometry
    cal_chs, cal_ch_numbers = _read_fine_cal(fine_cal)
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
1550, in _read_fine_cal
    with open(fine_cal, 'r') as fid:
IOError: [Errno 2] No such file or directory:
'/home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg/mnefun/data/sss_cal.dat'

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
@jyeatman are you sure you're on the right branch/ setup correctly?! You weren't encountering this before...

Eric89GXL:
It's because he's using `setup.py install` instead of `setup.py develop`. @kambysese you need to add any non-.py files here:

https://github.com/LABSN/mnefun/blob/master/setup.py#L39

jyeatman:
Can I switch to setup.py develop or do I need to do some in-installing
first?

On Wednesday, February 17, 2016, Eric Larson <notifications@github.com>
wrote:

> It's because he's using setup.py install instead of setup.py develop.
> @kambysese <https://github.com/kambysese> you need to add any non-.py
> files here:
>
> https://github.com/LABSN/mnefun/blob/master/setup.py#L39
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/LABSN/mnefun/pull/112#issuecomment-185436066>.
>


-- 
Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
As `package_data={'mnefun': ['run_sss.sh', 'data/sss_cal.dat']}`, or `...,sss_cal.dat` or what?

kambysese:
@jyeatman you'd have to re-install after cleaning up.

Eric89GXL:
I'd go with `rm -Rf /home/jyeatman/.local/lib/python2.7/site-packages/mnefun-0.1-py2.7.egg` just to be safe and then do `python setup.py develop --user`

Eric89GXL:
`package_data={'mnefun': ['run_sss.sh', op.join('data', 'sss_cal.dat'])}` should work

Eric89GXL:
oh and you'll need `ct_sparse.fif` in there too

kambysese:
@Eric89GXL 
> You mean the function run_sss_positions does not work properly?
No, It don't. Worst, is that it overwrites the `.pos` file with the smaller one generated from the split file.
I am going to have to bring over some of the things I did in the last commit over into `run_sss_positions` to handle the split files. Please, remind again why can't we use `_calculate_chpi_positions` from mne-python?

Eric89GXL:
> ￼it overwrites the .pos file with the smaller one generated from the split file.

Weird, I've used it for that purpose before. In any case, we should fix it while preserving the behavior that passing it a `.fif` file will automatically run any associated `-1.fif` files, too. So the while loop should still probably be there. The problem with the previous commits you had were that they made it so this behavior *changed*, so existing scripts I have that make use of this function will break.

Eric89GXL:
> remind again why can't we use _calculate_chpi_positions from mne-python

It needs to be extensively vetted (and corrected when it fails), and nobody has found time to do it yet. Even just identifying cases where it fails relative to MaxFilter would be a good first step, because it would give targets for improvement.

jyeatman:
A bit farther, so that's good. But now we actually have an error in
computations

  File "/home/jyeatman/git/BrainTools/projects/NLR_MEG/nlr_testmnefun.py",
line 120, in <module>
    gen_report=False
  File "/home/jyeatman/git/mnefun/mnefun/_mnefun.py", line 558, in
do_processing
    outs[ii] = func(p, subjects, run_indices)
  File "/home/jyeatman/git/mnefun/mnefun/_mnefun.py", line 727, in run_sss
    run_sss_localy(p, subjects, run_indices)
  File "/home/jyeatman/git/mnefun/mnefun/_mnefun.py", line 957, in
run_sss_localy
    head_pos=pos, regularize=reg)
  File "<string>", line 2, in maxwell_filter
  File "/home/jyeatman/git/mne-python/mne/utils.py", line 551, in verbose
    return function(*args, **kwargs)
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
406, in maxwell_filter
    n_positions, t_str)
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
583, in _do_tSSS
    t_proj = _overlap_projector(orig_in_data, resid, st_correlation)
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
1520, in _overlap_projector
    Q_int = linalg.qr(_orth_overwrite((data_int / n).T),
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
1498, in _orth_overwrite
    **check_disable)[:2]
  File
"/home/jyeatman/anaconda/lib/python2.7/site-packages/scipy/linalg/decomp_svd.py",
line 109, in svd
    raise LinAlgError("SVD did not converge")
numpy.linalg.linalg.LinAlgError: SVD did not converge

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
hahaha! congrats. What is the value of `params.tsss_dur`?

jyeatman:
params.tsss_dur = 4.

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
Could you try `6.`

jyeatman:
Tried 6. Same error


  File "/home/jyeatman/git/BrainTools/projects/NLR_MEG/nlr_testmnefun.py",
line 120, in <module>
    gen_report=False
  File "/home/jyeatman/git/mnefun/mnefun/_mnefun.py", line 558, in
do_processing
    outs[ii] = func(p, subjects, run_indices)
  File "/home/jyeatman/git/mnefun/mnefun/_mnefun.py", line 727, in run_sss
    run_sss_localy(p, subjects, run_indices)
  File "/home/jyeatman/git/mnefun/mnefun/_mnefun.py", line 957, in
run_sss_localy
    head_pos=pos, regularize=reg)
  File "<string>", line 2, in maxwell_filter
  File "/home/jyeatman/git/mne-python/mne/utils.py", line 551, in verbose
    return function(*args, **kwargs)
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
406, in maxwell_filter
    n_positions, t_str)
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
583, in _do_tSSS
    t_proj = _overlap_projector(orig_in_data, resid, st_correlation)
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
1520, in _overlap_projector
    Q_int = linalg.qr(_orth_overwrite((data_int / n).T),
  File "/home/jyeatman/git/mne-python/mne/preprocessing/maxwell.py", line
1498, in _orth_overwrite
    **check_disable)[:2]
  File
"/home/jyeatman/anaconda/lib/python2.7/site-packages/scipy/linalg/decomp_svd.py",
line 109, in svd
    raise LinAlgError("SVD did not converge")
numpy.linalg.linalg.LinAlgError: SVD did not converge


Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
The only thing I can say about this error is that I'm glad I am not the only one. Can you share the data file, I'm curious to see if I get the error on my system with the Anaconda's intel MKL optimization.

kambysese:
@Eric89GXL please review. Unless you have concerns about how & where splits are handled this is pretty close to done.

mnefun/_mnefun.py
```diff
@@ -829,29 +873,128 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
...
-    cmd = ['scp', '-P' + port] + fnames_in + [host + ':~/']
+    cmd = ['scp', '-P' + str(port)] + fnames_in + [host + ':~/']
     run_subprocess(cmd, stdout=None, stderr=None)
+    t0 = time()
+    remote_ins = ['~/' + op.basename(f) for f in fnames_in]
+    if len(remote_ins) > 1:
+        fnames_out = [op.basename(r)[:-4] + '.tmp' for r in remote_ins]
+        assert len(fnames_out) == len(fnames_in)
+    else:
+        fnames_out = [op.basename(fname_out)[:-4] + '.tmp']
```
Eric89GXL: do you need this conditional? It seems like it should work the same whether `len(remote_ins) == 1` or is `> 1`.

mnefun/_mnefun.py
```diff
@@ -829,29 +873,128 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
...
+    # concatenate hp pos file for split raw files if any
+    pos_data = []
+    with open(fname_out, 'w') as fo:
+        for f in fnames_out:
+            with open(op.join(pout, f)) as infile:
+                content = infile.readlines()
+                header = content[0]
+                qdata = content[1:]
+                pos_data.append(qdata)
+                os.remove(op.join(pout, f))
```
Eric89GXL: Only indent as many lines as necessary in the `with` context manager (in this case, I think just the `infile.readlines()` one). Especially this `os.remove` line, it should not be inside a context manager that has that file open.

mnefun/_mnefun.py
```diff
@@ -829,29 +873,128 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
...
-    cmd = ['ssh', '-p', port, host, 'rm -f %s %s %s'
+        print('  Cleaning up %s' % host)
+        cmd = ['ssh', '-p', str(port), host, 'rm -f %s %s %s'
            % (' '.join(remote_ins), remote_hp, remote_out)]
-    run_subprocess(cmd)
+        run_subprocess(cmd)
+
+    # concatenate hp pos file for split raw files if any
+    pos_data = []
+    with open(fname_out, 'w') as fo:
```
Eric89GXL: this `with` can be moved to the end, to only hold the `fo.writelines` line

mnefun/_mnefun.py
```diff
@@ -829,29 +873,128 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
...
+                content = infile.readlines()
+                header = content[0]
+                qdata = content[1:]
+                pos_data.append(qdata)
+                os.remove(op.join(pout, f))
+        pos_data.insert(0, header)
+        fo.writelines(''.join(str(j) for j in i) for i in pos_data)
+
+
+def run_sss_localy(p, subjects, run_indices):
```
Eric89GXL: locally, not localy

mnefun/_mnefun.py
```diff
@@ -829,29 +873,128 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
...
+    """Run SSS localy using maxwell filter in python
```
Eric89GXL: same here

mnefun/_mnefun.py
```diff
@@ -952,7 +1095,7 @@ def fix_eeg_files(p, subjects, structurals=None, dates=None, run_indices=None):
         names = [name for name in raw_names if op.isfile(name)]
         # noinspection PyPep8
         if structurals is not None and structurals[si] is not None and \
-                dates is not None:
+            dates is not None:
```
Eric89GXL: please restore the four spaces, this is a PEP8 violation (ambiguous indent)

setup.py
```diff
@@ -36,7 +36,7 @@
...
           platforms='any',
           packages=setuptools.find_packages(),
-          package_data={'mnefun': ['run_sss.sh']},
-          scripts=[
-              'bin/simulate_movement.py',
-          ])
+          package_data={'mnefun': ['run_sss.sh', os.path.join('data', 'sss_cal.dat'),
+                                   os.path.join('data', 'ct_sparse.fif')]},
+          scripts=['bin/simulate_movement.py'],
+          requires=['numpy'])
```
Eric89GXL: Please remove the `numpy` requirement here, we don't want `python` trying to use `easy_install` or `pip` or anything else to install it. There are additional requirements anyway, and they match what `mne`'s requirements are.

Eric89GXL:
I assume you've tested this and it works well on some dataset?

Code LGTM other than minor comments

kambysese:
Will address comments before Mon., Tested this with e.g. data and my own that included split files. i.e., pos arrays were concatenated as expected.

kambysese:
ready for merge

mnefun/_mnefun.py
```diff
@@ -829,29 +873,125 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
...
+        with open(op.join(pout, f)) as infile:
+            content = infile.readlines()
+        header = content[0]
+        qdata = content[1:]
+        pos_data.append(qdata)
+        os.remove(op.join(pout, f))
+    pos_data.insert(0, header)
+
+    with open(fname_out, 'w') as fo:
+        fo.writelines(''.join(str(j) for j in i) for i in pos_data)
```
Eric89GXL: Actually instead of messing with all this writing, you should be able to use `mne.chpi.read_head_quats` to get arrays, concatenate them along axis 0 with `np.concatenate(..., axis=0)` and then write the result with `mne.chpi.write_head_quats`

mnefun/_mnefun.py
```diff
@@ -1458,7 +1598,8 @@ def _raw_LRFCP(raw_names, sfreq, l_freq, h_freq, n_jobs, n_jobs_resample,
         print('    Loading and filtering %d files.' % len(raw_names))
     raw = list()
     for rn in raw_names:
-        r = Raw(rn, preload=True)
+        with warnings.catch_warnings(record=True):
+            r = Raw(rn, preload=True, allow_maxshield=allow_maxshield)
```
Eric89GXL: you can change this to `allow_maxshield='yes'` and then you don't need the `with` context manager (new feature in mne-python)

setup.py
```diff
@@ -36,7 +36,6 @@
           classifiers=[],
           platforms='any',
           packages=setuptools.find_packages(),
-          package_data={'mnefun': ['run_sss.sh']},
-          scripts=[
-              'bin/simulate_movement.py',
-          ])
+          package_data={'mnefun': ['run_sss.sh', os.path.join('data', 'sss_cal.dat'),
+                                   os.path.join('data', 'ct_sparse.fif')]},
+          scripts=['bin/simulate_movement.py'])
```
Eric89GXL: PEP8 don't forget the newline

Eric89GXL:
Other than the last couple minor tweaks, LGTM. Thanks for sticking with it.

mnefun/_mnefun.py
kambysese: I assume you mean `read_head_pos` & `write_head_pos`?

mnefun/_mnefun.py
kambysese: Still getting a warning with `allow_maxshield='yes'`. I updated to upstream/master yesterday.

kambysese: NVMND

kambysese:
Please make this go away.

Eric89GXL:
Thanks @kambysese

PR #116, opened 2016-03-10
WIP: second level preliminary group analysis

kambysese:
@Eric89GXL very rough sketch of what I hope to do beyond current endpoint of mnefun, see [here](https://gist.github.com/kambysese/73a079256d64d218276a) LMKWYT about (a) how I've modified the mnefun directory tree and (b) general thoughts. 
It would also be useful to here what other LABSN users think.


mnefun/stats/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Functions for statistical analysis"""
+
+from mnefun import _paths
```
Eric89GXL: use relative paths, not absolute -- and actually this shouldn't be imported in `__init__.py` at all, but rather in whatever function actually needs it

Eric89GXL:
Seems fine to me. It could make for at the very least some really nice first-pass sensor space comparisons for adult data. But I assume you have plans to use it for infant data, which should work, too.

kambysese:
That's what I want initially i.e., first pass sensor domain feel for data and eventually some sort of overall source domain comparisons if at all feasible. Agnostic, as far as data type, infant, adult, monkey :)

PR #117, opened 2016-03-11
FIX: minor assertion to cover args

kambysese:
@Eric89GXL minor touch up if you're ok with syntax.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -214,7 +214,7 @@ class Params(Frozen):
         ovwrlapping intersecting inner/outer signals during spatiotemporal SSS.
         Default is .98 however a smaller value of .9 is recommended for infant/
         child data.
-    trans_to : str | None
+    trans_to : str | array-like, (3,) | None
         The destination location for the head. Can be ``None``, which
         will not change the head position, or a string path to a FIF file
         containing a MEG device<->head transformation. Default is median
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -949,7 +949,6 @@ def run_sss_locally(p, subjects, run_indices):
             raw.fix_mag_coil_types()
             raw = filter_chpi(raw)
 
-            assert isinstance(p.trans_to, string_types)
             if p.trans_to is 'median':
```
Eric89GXL: crap, not sure how this got in, string comparisons should be `==` not `is`.

Also, why remove this assertion?

kambysese: wdym how it got in? most likely I got it in, you have a check for this or what? 

I rm it bc (i) `trans_to` may also be an `np.array` (ii) thought the `if` statement should be enough, no? I can go with

```
assert isinstance(p.trans_to, (string_types, np.ndarray)) 
             if p.trans_to == 'median':
```
 ?

Eric89GXL: I mean I missed it in review...

Don't worry about putting the assert back. I was initially thinking that
it's bad to keep allowing options that only work with the Python algorithm,
but I think it's okay.

mnefun/_mnefun.py
```diff
@@ -949,7 +949,7 @@ def run_sss_locally(p, subjects, run_indices):
             raw.fix_mag_coil_types()
             raw = filter_chpi(raw)
 
-            assert isinstance(p.trans_to, string_types)
+            assert type(p.trans_to) in (str, np.ndarray, type(None))
```
kambysese: @Eric89GXL not sure if this is pretty or acceptable but seems to work.

Eric89GXL: `assert` statements can be skipped by Python when it's executed with `-O`. What you probably want in any case is `isinstance(p.trans_to, (string_types, np.ndarray, type(None)))`

kambysese:
Thanks @Eric89GXL ready for merge...

Eric89GXL:
Thanks @kambysese

PR #118, opened 2016-03-14
Avg pos

kambysese:
@Eric89GXL as per [this](https://github.com/LABSN/larsoner/issues/1) I think this is what we want for time weighted mean head pos. LMWYT.
The testing is something I started for my own edumecation feel free to 86.


mnefun/_mnefun.py
```diff
@@ -2122,4 +2122,22 @@ def _headpos(p, file_in, file_out):
...
+    """Helper for computing average head transformation matrix
+    without rotation from cHPI data"""
+    assert pos.shape[1] == 10
+    t = pos[:, 0]
+    # q4:q6 cols from Maxfilter hp file
+    trans = pos[:, 4:7]
+    # time weighting factor as col vector
+    frac = np.diff(t)[:, np.newaxis] / (t[-1] - t[0])
+    assert frac.size == trans[:-1].shape[0]
+    if mode == 'mean':
```
Eric89GXL: why the conditional? if `mode != 'mean'` then `p0` won't even exist...

kambysese: For future if & when we want an option between mean or median hp.

Eric89GXL: Then add an else clause that raises an error

PR #120, opened 2016-04-05
WIP: Add compute_auc

larsoner:
Should be good for @mmittag


kambysese:
Is this for `.fif` type dipole files? perhaps @mmittag could elaborate the docstring? otherwise LGTM.

PR #123, opened 2016-04-29
FIX: Minor fixes

Eric89GXL:
@kambysese fixes a few things that were broken recently due to `mnefun` Maxwell-filtering changes as well as `mne-python` updates. Ready for review from my end.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -345,7 +345,7 @@ class Params(Frozen):
         # This is used by fix_eeg_channels to fix original files
         self.raw_fif_tag = '_raw.fif'
         # SSS denoising params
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -344,7 +344,7 @@ def __init__(self, tmin=None, tmax=None, t_adjust=0, bmin=-0.2, bmax=0.0,
         # This is used by fix_eeg_channels to fix original files
         self.raw_fif_tag = '_raw.fif'
         # SSS denoising params
-        self.sss_type = 'python'
+        self.sss_type = 'maxfilter'
```
Eric89GXL: e.g. we still want this as the default for backward compatibility I think

PR #126, opened 2016-05-10
Groupstats

jyeatman:
@Eric89GXL @wronk Chances are this is either already in mnefun or you guys will have a much better suggestion for how I should go about this. But I wanted to start automating doing some simple group level statistics (first at the sensor level then in source space), so I thought I would moch up a function and then start the conversation about the best way to go about this.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -922,6 +922,26 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
     pos_data = np.concatenate(np.array(data))
     write_head_pos(fname_out, pos_data)
 
+def group_analysis(p, subjects, group, stats):
+    """Generate group level statistics
+
+    Since we know all the subjects and the data structure we can automatically
+    compute some statistics with minimal information. This function Will
+    compute different statistics on the data and save it into the grp_avg
+    folder
+
+    Inputs:
+    p        - params structure
+    subjects -
+    groups   - list of group names (or numbers), where each list entry denotes
+               the group that a subject belongs to.
+    stats    - List of statistics to compute on the group. Options are 'mean',
+               'ttest', etc
+    """
+    for si, subj in enumerate(subjects):
+        raw_dir = op.join(p.work_dir, subj, p.raw_dir)
+        sss_dir = op.join(p.work_dir, subj, p.sss_dir)
+        evoked_dir = op.join(p.work_dir, subj, p.inverse_dir)
 
 def run_sss_locally(p, subjects, run_indices):
     """Run SSS locally using maxwell filter in python


Eric89GXL:
I've wanted to do something like this for a while, but always get stuck at the planning stages. What tests should we do that are useful in 90% of cases (or is that too strict a standard)? Some studies contrast within-groups, some between. Do we want a simple mass univariate measure?

Maybe if we allow only for mass univariate (uncorrected for now) 1-sample (across conditions) or 2-sample (across groups) we could cover a lot of good use cases.

I've also been meaning to revamp the event coding system, since I suspect we could get away with something *much* more transparent now that mne-python has what we call hierarchical event IDs, but that's probably for another discussion.

Eric89GXL:
cc @kambysese what do you think would be most useful for your studies? I suspect having the option of some simple mass-univariate within- or between-group comparisons would cover a lot of it.

jyeatman:
Yeah, I think that is right on! My feeling is that it would be incredibly
helpful to have anything even just calculating the mean. If the code is
written clearly then it can serve as a template for people to write their
own more custom statistics. And at the very least it will help for rapid
debugging. For example even being able to quickly look at the group mean
for each condition would make it easy to quickly explore datasets and test
the effect of different preprocessing parameter combinations

On Wednesday, May 11, 2016, Eric Larson <notifications@github.com> wrote:

> cc @kambysese <https://github.com/kambysese> what do you think would be
> most useful for your studies? I suspect having the option of some simple
> mass-univariate within- or between-group comparisons would cover a lot of
> it.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/LABSN/mnefun/pull/126#issuecomment-218455684>
>


-- 
Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
@Eric89GXL we've thrown this idea back & forth several times now. Personally I've been unable to get too far along passed the "planning stages" WRT to inferential tests; even across my own projects much less defining agnostic routines for other users. That said, I do find that I replicate code when it comes to group level analyses. 
Initially, to be clear I think we should prioritize and begin with descriptive over inferential routines and use `params.analyses` as a starting point. That should provide a better opportunity to id subsequent inferential routines for  measures of e.g. magnitude or latency in both domains (sensor and source). I say this because for a majority of evoked response studies repeated measures are defined based off descriptives.

kambysese:
@Eric89GXL if you're considering revamping the event coding scheme, I wonder it it makes any sense to separate it into e.g. it's own stand alone class type from `params`?

jyeatman:
A first pass could be to output sensor level maps of the mean+se for each
condition and each group.

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

kambysese:
What do you mean by sensor level maps? Take a look at [this](https://github.com/kambysese/ilabs/blob/master/bad_baby/Bad%20Tone.ipynb) and LMKWYT. Here I am mainly trying to get a better idea of individual data quality and group level ERFs for groups. Somethings can be readily applied to group level procedures, but eg. the sensor level tests operate on epochs and I don't know of a legit means of extending that into group level "sensor level maps" @Eric89GXL thoughts?

jyeatman:
That plus maps of ERPs+CIs would be good

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

Eric89GXL:
> if you're considering revamping the event coding scheme, I wonder it it makes any sense to separate it into e.g. it's own stand alone class type from params?

I'd rather not make a class unless we have to. I think we can get away with some simpler coding, but let's deal with that in a separate issue so we don't derail this one too far.

> Take a look at this and LMKWYT.

There is a lot of data and plots in there, what specifically did you have in mind?

> A first pass could be to output sensor level maps of the mean+se for each condition and each group.

Sounds good to me. I think @staulu has been asking about this previously, too.

> the sensor level tests operate on epochs and I don't know of a legit means of extending that into group level "sensor level maps"

I think you're right that for the first steps let's not worry about stats. Let's do individual stuff (as above) plus grand averages. Getting that done will be useful enough, and probably require some discussion about API.

jyeatman:
I agree with Eric on all that. Maybe we shouldn't be thinking about "doing
stats", but instead generating useful visualizations of the results.

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

larsoner:
Closing for now, but we can brainstorm ideas for additional functionality moving forward

PR #128, opened 2016-05-24
Minor fix to squash deprecation error

kambysese:
@Eric89GXL should I be concerned about backward compatibility here?


Eric89GXL:
I think that old scripts should continue to work as they did before, assuming you have an up-to-date mne-python. So this is fine

PR #130, opened 2016-05-25
Safer file name handling during fetch

kambysese:
@Eric89GXL this change works with the example ds. LMKWYT. If go for merge then I can test with Alexis' dataset. Also, a couple of automated PEP8 fixes snuck in. LMK if you have a problem with them.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -40,6 +40,7 @@ from mne.preprocessing.maxwell import (maxwell_filter,
                                        _get_mf_picks, _prep_mf_coils,
                                        _check_regularize,
                                        _regularize)
+
 try:
     # Experimental version
     from mne.preprocessing.maxwell import _prep_regularize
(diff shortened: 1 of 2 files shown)


mnefun/_mnefun.py
```diff
@@ -600,7 +601,10 @@ def fetch_raw_files(p, subjects, run_indices):
                                   p.acq_ssh, finder])[0]
         remote_fnames = [x.strip() for x in stdout_.splitlines()]
         assert all(fname.startswith(p.acq_dir) for fname in remote_fnames)
-        remote_fnames = [fname[len(p.acq_dir) + 1:] for fname in remote_fnames]
+        p_dir, _ = op.split(remote_fnames[0])
```
Eric89GXL: this assumes that the directory is only two parts like `/a/b` and not `/a/b/c` right? That's not a general/safe enough assumption

Eric89GXL: ...oh never mind, that's not what it does :)

Eric89GXL:
Let's see if it works...

PR #132, opened 2016-06-03
FIX: Minor fix to filename

Eric89GXL:
@kambysese your recent changes broke my pulls :( Can you see if this works for you?


kambysese:
Sorry about that. This works for me. I can only assume it will work for the error I ran into for Alexis' data, I never had a chance to try it out b/c her machine is water damaged. We'll cross that bridge later.

Eric89GXL:
Thanks for the look @kambysese

PR #133, opened 2016-06-15
FIX: Minor fixes

Eric89GXL:
@wronk can you see if this fixes your problem with missing channels?

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -356,6 +356,7 @@ class Params(Frozen):
         self.st_correlation = .98
         self.sss_origin = 'auto'
         self.sss_regularize = 'in'
+        self.filter_chpi = True
         # boolean for whether data set(s) have an individual mri
         self.on_process = None
         # Use more than EXTRA points to fit headshape
@@ -740,7 +741,7 @@ def run_sss(p, subjects, run_indices):
     """Run SSS preprocessing remotely (only designed for *nix platforms) or
     locally using Maxwell filtering in mne-python"""
     if p.sss_type == 'python':
-        print(' Applying SSS locally using mne-python')
+        print('  Applying SSS locally using mne-python')
         run_sss_locally(p, subjects, run_indices)
     else:
         for si, subj in enumerate(subjects):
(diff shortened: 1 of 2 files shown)


mnefun/_mnefun.py
```diff
@@ -995,19 +998,47 @@ def run_sss_locally(p, subjects, run_indices):
...
-                    raise NameError('File not found (' + r + ')')
-                raw = Raw(r, preload=True, allow_maxshield=True)
-                raw.load_bad_channels(prebad_file, force=True)
-                # apply maxwell filter
-                raw_sss = maxwell_filter(
-                    raw, int_order=p.int_order, ext_order=p.ext_order,
-                    calibration=cal_file, cross_talk=ct_file,
-                    st_correlation=p.st_correlation, st_duration=st_duration,
-                    destination=None, coord_frame='meg')
-                raw_sss.save(o, overwrite=True, buffer_size_sec=None)
```
Eric89GXL: @kambysese this loop was at the wrong level so I think these files were being processed many times more than they needed to be

kambysese: Sorry about that!

wronk:
Looks like it worked, thanks

mnefun/_mnefun.py
```diff
@@ -862,7 +878,8 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
             pass
 
 
-def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22):
+def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22,
+                      prefix='  '):
```
wronk: `port` and `prefix` missing from docstring

mnefun/_mnefun.py
```diff
@@ -1567,7 +1616,7 @@ def gen_covariances(p, subjects, run_indices):
             epochs = Epochs(raw, events, event_id=None, tmin=p.bmin,
                             tmax=p.bmax, baseline=(None, None), proj=False,
                             reject=use_reject, flat=use_flat, preload=True)
-            epochs.pick_types(meg=True, eeg=True, exclude=[])
+            epochs.pick_types(meg=True, eeg=True, exclude=())
```
wronk: ??

wronk:
Couple minor things, but LGTM other than those

mnefun/_mnefun.py
Eric89GXL: It won't make much of a difference. I slightly prefer tuples for this sort of thing since they are immutable, but I'm not usually very careful about it. Not really sure why I changed this one...

Eric89GXL:
Fixed, thanks @wronk

PR #134, opened 2016-06-17
refactor into mnefun / bin

jjnurminen:
I had to modify sys.path in the beginning of the executable script to import mnefun. Let me know if you prefer some other solution.

### DIFF
--- a/bin/chpi_snr.py
+++ b/bin/chpi_snr.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Plot signal-to-noise of continuous HPI coils as a function of time.
+Works by fitting a general linear model (HPI freqs, line freqs, DC, slope) to
+the data, and comparing estimated HPI powers with the residual (=variance
+unexplained by the model).
+Window length for SNR estimates can be specified on the command line.
+Longer windows will by nature include more low frequencies and thus have
+larger residual variance (lower SNR).
+
+Tested with Python 2.7, MNE 0.11.0
+@author: jussi (jnu@iki.fi)
+"""
+
+import sys
+import os
+sys.path.append(os.path.join(os.path.dirname(__file__), os.pardir))
+import mne
+import argparse
+from mnefun import chpi_snr_raw
+
+
+def run():
+
+    # default parameters
+    default_winlen = 1  # window length, seconds
+    default_nharm = 2  # number of line harmonics to include
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('fiff_file', help='Name of raw fiff file')
+    parser.add_argument('--winlen', type=float, default=default_winlen,
+                        help='Buffer length for SNR estimates (s)')
+    parser.add_argument('--nharm', type=int, default=default_nharm,
+                        choices=[0, 1, 2, 3, 4], help='Number of line'
+                        'frequency harmonics to include')
+    parser.add_argument('--stop', type=float, metavar='t', default=None,
+                        help='Process only first t seconds')
+    args = parser.parse_args()
+
+    raw = mne.io.Raw(args.fiff_file, allow_maxshield=True)
+    chpi_snr_raw(raw, args.winlen, args.nharm, args.stop)
+
+
+if __name__ == '__main__':
+    run()
--- a/bin/plot_chpi_snr.py
+++ b/bin/plot_chpi_snr.py
@@ -31,12 +31,10 @@ def run():
     parser.add_argument('--nharm', type=int, default=default_nharm,
                         choices=[0, 1, 2, 3, 4], help='Number of line'
                         'frequency harmonics to include')
-    parser.add_argument('--stop', type=float, metavar='t', default=None,
-                        help='Process only first t seconds')
     args = parser.parse_args()
 
     raw = mne.io.Raw(args.fiff_file, allow_maxshield='yes')
-    plot_chpi_snr_raw(raw, args.winlen, args.nharm, args.stop)
+    plot_chpi_snr_raw(raw, args.winlen, args.nharm)
 
 
 if __name__ == '__main__':
--- a/bin/plot_chpi_snr.py
+++ b/bin/plot_chpi_snr.py
@@ -8,9 +8,6 @@ unexplained by the model).
 Window length for SNR estimates can be specified on the command line.
 Longer windows will by nature include more low frequencies and thus have
 larger residual variance (lower SNR).
-
-Tested with Python 2.7, MNE 0.11.0
-@author: jussi (jnu@iki.fi)
 """
 
 import sys
@@ -18,7 +15,7 @@ import os
 sys.path.append(os.path.join(os.path.dirname(__file__), os.pardir))
 import mne
 import argparse
-from mnefun import chpi_snr_raw
+from mnefun import plot_chpi_snr_raw
 
 
 def run():
@@ -38,8 +35,8 @@ def run():
                         help='Process only first t seconds')
     args = parser.parse_args()
 
-    raw = mne.io.Raw(args.fiff_file, allow_maxshield=True)
-    chpi_snr_raw(raw, args.winlen, args.nharm, args.stop)
+    raw = mne.io.Raw(args.fiff_file, allow_maxshield='yes')
+    plot_chpi_snr_raw(raw, args.winlen, args.nharm, args.stop)
 
 
 if __name__ == '__main__':
--- a/bin/plot_chpi_snr.py
+++ b/bin/plot_chpi_snr.py
@@ -5,6 +5,7 @@ Plot signal-to-noise of continuous HPI coils as a function of time.
 Works by fitting a general linear model (HPI freqs, line freqs, DC, slope) to
 the data, and comparing estimated HPI powers with the residual (=variance
 unexplained by the model).
+
 Window length for SNR estimates can be specified on the command line.
 Longer windows will by nature include more low frequencies and thus have
 larger residual variance (lower SNR).
@@ -12,7 +13,6 @@ larger residual variance (lower SNR).
 
 import sys
 import os
-sys.path.append(os.path.join(os.path.dirname(__file__), os.pardir))
 import mne
 import argparse
 from mnefun import plot_chpi_snr_raw
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -2434,7 +2434,7 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
     # SNR plots for gradiometers and magnetometers
     ax = axs[0]
     lines1 = ax.plot(tvec, 10*np.log10(snr_avg_grad.transpose()))
-    ax.set(ylabel='SNR (dB)', xlabel='Time (s)')
+    ax.set(ylabel='SNR (dB)')
     ax.set_title('Mean cHPI power / mean residual variance, gradiometers',
                  fontsize=title_fontsize)
     ax = axs[1]
@@ -2452,10 +2452,10 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
     cls = plt.get_cmap('plasma')(np.linspace(0., 0.7, len(pick_meg)))
     ax.set_prop_cycle(color=cls)
     ax.semilogy(tvec, resid_vars[pick_grad, :].transpose(), alpha=.4)
-    ax.set(ylabel='Var. (T/m)$^2$')
+    ax.set(ylabel='Var. (T/m)$^2$', xlabel='Time (s)')
     ax.set_title('Residual (unexplained) variance, all gradiometer channels',
                  fontsize=title_fontsize)
-    tight_layout()  # from mne.viz
+    tight_layout(pad=.5, w_pad=.1, h_pad=.1)  # from mne.viz
     # tight_layout will screw these up
     ax = axs[0]
     box = ax.get_position()
@@ -2463,21 +2463,21 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
     # order curve legends according to mean of data
     sind = np.argsort(snr_avg_grad.mean(axis=1))[::-1]
     ax.legend(np.array(lines1)[sind], np.array(cfreqs_legend)[sind],
-              prop={'size': legend_fontsize}, bbox_to_anchor=(1, 0.5, ),
+              prop={'size': legend_fontsize}, bbox_to_anchor=(1.02, 0.5, ),
               loc='center left')
     ax = axs[1]
     box = ax.get_position()
     ax.set_position([box.x0, box.y0, box.width * 0.8, box.height])
     sind = np.argsort(snr_avg_mag.mean(axis=1))[::-1]
     ax.legend(np.array(lines2)[sind], np.array(cfreqs_legend)[sind],
-              prop={'size': legend_fontsize}, bbox_to_anchor=(1, 0.5, ),
+              prop={'size': legend_fontsize}, bbox_to_anchor=(1.02, 0.5, ),
               loc='center left')
     ax = axs[2]
     box = ax.get_position()
     ax.set_position([box.x0, box.y0, box.width * 0.8, box.height])
     sind = np.argsort(hpi_pow_grad.mean(axis=1))[::-1]
     ax.legend(np.array(lines3)[sind], np.array(cfreqs_legend)[sind],
-              prop={'size': legend_fontsize}, bbox_to_anchor=(1, 0.5, ),
+              prop={'size': legend_fontsize}, bbox_to_anchor=(1.02, 0.5, ),
               loc='center left')
     ax = axs[3]
     box = ax.get_position()
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -2373,7 +2373,9 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
 
     # plotting parameters
     legend_fontsize = 9
-    title_fontsize = 12
+    title_fontsize = 10
+    tick_fontsize = 10
+    label_fontsize = 10
 
     # get some info from fiff
     sfreq = raw.info['sfreq']
@@ -2435,27 +2437,36 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
     ax = axs[0]
     lines1 = ax.plot(tvec, 10*np.log10(snr_avg_grad.transpose()))
     ax.set(ylabel='SNR (dB)')
+    ax.yaxis.label.set_fontsize(label_fontsize)
     ax.set_title('Mean cHPI power / mean residual variance, gradiometers',
                  fontsize=title_fontsize)
+    ax.tick_params(axis='both', which='major', labelsize=tick_fontsize)
     ax = axs[1]
     lines2 = ax.plot(tvec, 10*np.log10(snr_avg_mag.transpose()))
     ax.set(ylabel='SNR (dB)')
+    ax.yaxis.label.set_fontsize(label_fontsize)
     ax.set_title('Mean cHPI power / mean residual variance, magnetometers',
                  fontsize=title_fontsize)
+    ax.tick_params(axis='both', which='major', labelsize=tick_fontsize)
     ax = axs[2]
     lines3 = ax.plot(tvec, hpi_pow_grad.transpose())
     ax.set(ylabel='Power (T/m)$^2$')
+    ax.yaxis.label.set_fontsize(label_fontsize)
     ax.set_title('Mean cHPI power, gradiometers',
                  fontsize=title_fontsize)
+    ax.tick_params(axis='both', which='major', labelsize=tick_fontsize)
     # residual (unexplained) variance as function of time
     ax = axs[3]
     cls = plt.get_cmap('plasma')(np.linspace(0., 0.7, len(pick_meg)))
     ax.set_prop_cycle(color=cls)
     ax.semilogy(tvec, resid_vars[pick_grad, :].transpose(), alpha=.4)
     ax.set(ylabel='Var. (T/m)$^2$', xlabel='Time (s)')
+    ax.xaxis.label.set_fontsize(label_fontsize)
+    ax.yaxis.label.set_fontsize(label_fontsize)
     ax.set_title('Residual (unexplained) variance, all gradiometer channels',
                  fontsize=title_fontsize)
-    tight_layout(pad=.5, w_pad=.1, h_pad=.1)  # from mne.viz
+    ax.tick_params(axis='both', which='major', labelsize=tick_fontsize)
+    tight_layout(pad=.5, w_pad=.1, h_pad=.2)  # from mne.viz
     # tight_layout will screw these up
     ax = axs[0]
     box = ax.get_position()
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -2339,7 +2339,7 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
 
     Parameters
     ----------
-    win_length : int
+    win_length : float
         Length of window to use for SNR estimates (seconds). A longer window
         will naturally include more low frequency power, resulting in lower
         SNR.
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -2348,7 +2348,7 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
 
     Returns
     -------
-    fig : Instance of matplotlib.figure.Figure
+    fig : instance of matplotlib.figure.Figure
         cHPI SNR as function of time, residual variance.
 
     Notes
@@ -2357,6 +2357,7 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
     each data window. The cHPI power obtained from the model is then divided
     by the residual variance (variance of signal unexplained by the model) to
     obtain the SNR.
+
     The SNR may decrease either due to decrease of cHPI amplitudes (e.g.
     head moving away from the helmet), or due to increase in the residual
     variance. In case of broadband interference that overlaps with the cHPI
@@ -2382,7 +2383,7 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
     if buflen <= 0:
         raise ValueError('Window length should be >0')
     (cfreqs, _, _, _, _) = _get_hpi_info(raw.info)
-    print('\nNominal cHPI frequencies: %s Hz' % cfreqs)
+    print('Nominal cHPI frequencies: %s Hz' % cfreqs)
     print('Sampling frequency: %s Hz' % sfreq)
     print('Using line freqs: %s Hz' % linefreqs)
     print('Using buffers of %s samples = %s seconds\n'
@@ -2423,9 +2424,9 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics):
         # sinusoidal of amplitude A has power of A**2/2
         hpi_pow = (coeffs_hpi[0::2, :]**2 + coeffs_hpi[1::2, :]**2)/2
         # divide average HPI power by average variance
-        snr_avg_grad[:, ind] = (hpi_pow)[:, pick_grad].mean(1) / \
+        snr_avg_grad[:, ind] = hpi_pow[:, pick_grad].mean(1) / \
             resid_vars[pick_grad, ind].mean()
-        snr_avg_mag[:, ind] = (hpi_pow)[:, pick_mag].mean(1) / \
+        snr_avg_mag[:, ind] = hpi_pow[:, pick_mag].mean(1) / \
             resid_vars[pick_mag, ind].mean()
 
     cfreqs_legend = ['%s Hz' % fre for fre in cfreqs]
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -16,6 +16,7 @@ from collections import Counter
 from time import time
 
 import numpy as np
+import scipy
 from scipy import io as spio
 import matplotlib.pyplot as plt
 from numpy.testing import assert_allclose
@@ -50,7 +51,8 @@ from mne.bem import _check_origin
 from mne.minimum_norm import make_inverse_operator
 from mne.label import read_label
 from mne.epochs import combine_event_ids
-from mne.chpi import (filter_chpi, read_head_pos, write_head_pos)
+from mne.chpi import (filter_chpi, read_head_pos, write_head_pos,
+                      _get_hpi_info)
 
 try:
     from mne.chpi import quat_to_rot, rot_to_quat
@@ -2332,26 +2334,28 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
     return fig
 
 
-def chpi_snr_raw(raw, winlen, nharm, stop=None):
-    """ Compute and plot cHPI SNR from raw data.
+def plot_chpi_snr_raw(raw, winlen, nharm, stop=None):
+    """Compute and plot cHPI SNR from raw data
 
     Parameters
     ----------
-
     winlen : int
         Length of window to use. Longer windows will naturally include more
         low frequency power, resulting in lower SNR.
-
     nharm : int
         Number of line frequency harmonics to include in the model.
-
     stop : int
         If specified, stop before the end of file (seconds)
+
+    Returns
+    -------
+    fig : Instance of matplotlib.figure.Figure
+        cHPI SNR as function of time, residual variance.
     """
 
     # plotting parameters
-    legend_fontsize = 12
-    legend_hspace = 30  # % of horizontal space to reserve for legend
+    legend_fontsize = 10
+    legend_hspace = 25  # % of horizontal axis to reserve for legend
 
     # get some info from fiff
     sfreq = raw.info['sfreq']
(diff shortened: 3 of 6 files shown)


jjnurminen:
As a bonus, code now passes flake8 checks...

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    if buflen <= 0:
+        raise ValueError('Window length should be >0')
+    cfreqs = []
+    if (len(raw.info['hpi_meas']) > 0 and
+            'coil_freq' in raw.info['hpi_meas'][0]['hpi_coils'][0]):
+        for coil in raw.info['hpi_meas'][0]['hpi_coils']:
+            cfreqs.append(coil['coil_freq'][0])
+    else:
+        raise ValueError('Cannot determine cHPI frequencies'
+                         'from raw data info')
```
Eric89GXL: To make this DRY you should use [_get_hpi_info] (https://github.com/mne-tools/mne-python/blob/master/mne/chpi.py#L122)

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+
+    winlen : int
+        Length of window to use. Longer windows will naturally include more
+        low frequency power, resulting in lower SNR.
+
+    nharm : int
+        Number of line frequency harmonics to include in the model.
+
+    stop : int
+        If specified, stop before the end of file (seconds)
```
Eric89GXL: See how other docstrings are formatted in MNE-Python and try to follow it as closely as possible. It should probably also return the Matplotilb figure (see other MNE-Python viz functions)

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+            cfreqs.append(coil['coil_freq'][0])
+    else:
+        raise ValueError('Cannot determine cHPI frequencies'
+                         'from raw data info')
+    print('\nNominal cHPI frequencies: ', cfreqs, ' Hz')
+    print('Sampling frequency:', sfreq, 'Hz')
+    print('Using line freqs:', linefreqs, ' Hz')
+    print('Using buffers of', buflen, 'samples =', buflen/sfreq, 'seconds\n')
+
+    raw.info['bads'] = []  # make sure to load all channels
```
Eric89GXL: don't do this, it modifies `raw` -- if someone uses it afterward this will be bad

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    pick_meg = pick_types(raw.info, meg=True)
```
Eric89GXL: ...instead use `exclude=()` (the default is `exclude='bads'` which I think is why you had the `bads` line above?)

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    pick_mag = pick_types(raw.info, meg='mag')
+    pick_grad = pick_types(raw.info, meg='grad')
+    nchan = len(pick_meg)
+
+    # create general linear model for the data
+    t = np.linspace(0, buflen/sfreq, endpoint=False, num=buflen)
```
Eric89GXL: floating point arithmetic is not exact, better to use integer:
```
t = np.arange(buflen) / float(sfreq)
```

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    model = np.c_[t, np.ones(t.shape)]  # model slope and DC
+    # add sine and cosine term for each freq
+    for f in list(linefreqs) + cfreqs:
+        model = np.c_[model, np.cos(2*np.pi*f*t), np.sin(2*np.pi*f*t)]
+    inv_model = np.linalg.pinv(model)
```
Eric89GXL: It's better (usually) to use `scipy.linalg` instead

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+
+    # loop through MEG data, fit linear model and compute SNR at each window
+    if stop:
+        stop = int(stop*sfreq)
+    else:
+        stop = raw.n_times
+    # drop last buffer to avoid overrun
+    bufs = range(0, int(stop), buflen)[:-1]
```
Eric89GXL: the `int` cast here will be redundant

Eric89GXL: ... and `range` on Py3k is not a list but a generator, so it's better to use `np.arange` instead

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    tvec = np.array(bufs)/sfreq
+    snr_avg_grad = np.zeros([len(cfreqs), len(bufs)])
+    snr_avg_mag = np.zeros([len(cfreqs), len(bufs)])
+    resid_vars = np.zeros([nchan, len(bufs)])
+    for ind, buf0 in enumerate(bufs):
+        megbuf = raw[pick_meg, buf0:buf0+buflen][0].transpose()
```
Eric89GXL: can just do `raw[...].T` instead of `.transpose()`

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+        coeffs = np.dot(inv_model, megbuf)
+        coeffs_hpi = coeffs[2+2*len(linefreqs):]
+        resid_vars[:, ind] = np.var(megbuf-np.dot(model, coeffs), 0)
+        # get total HPI amplitudes by combining sine and cosine terms
+        hpi_amps = np.sqrt(coeffs_hpi[0::2, :]**2 + coeffs_hpi[1::2, :]**2)
+        # divide average HPI power by average variance
+        snr_avg_grad[:, ind] = np.divide((hpi_amps**2/2)[:, pick_grad].mean(1),
```
Eric89GXL: why `np.dividie` instead of `(... / ...)`?

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+                                         resid_vars[pick_grad, ind].mean())
+        snr_avg_mag[:, ind] = np.divide((hpi_amps**2/2)[:, pick_mag].mean(1),
+                                        resid_vars[pick_mag, ind].mean())
+
+    plt.figure()
+    cfreqs_legend = [str(fre)+' Hz' for fre in cfreqs]
```
Eric89GXL: A more standard way would be `['%s Hz' % fre for fre in cfreqs]`

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+        resid_vars[:, ind] = np.var(megbuf-np.dot(model, coeffs), 0)
+        # get total HPI amplitudes by combining sine and cosine terms
+        hpi_amps = np.sqrt(coeffs_hpi[0::2, :]**2 + coeffs_hpi[1::2, :]**2)
+        # divide average HPI power by average variance
+        snr_avg_grad[:, ind] = np.divide((hpi_amps**2/2)[:, pick_grad].mean(1),
+                                         resid_vars[pick_grad, ind].mean())
+        snr_avg_mag[:, ind] = np.divide((hpi_amps**2/2)[:, pick_mag].mean(1),
+                                        resid_vars[pick_mag, ind].mean())
+
+    plt.figure()
```
Eric89GXL: fig = plt.figure()

Eric89GXL: ...and actually it's better to do `fig, ax = plt.subplots()` and use methods instead of the global namespace functions

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    cfreqs_legend = [str(fre)+' Hz' for fre in cfreqs]
+
+    # order curve legends according to mean of data
+    sind = np.argsort(snr_avg_grad.mean(axis=1))[::-1]
+    lines1 = plt.plot(tvec, 10*np.log10(snr_avg_grad.transpose()))
```
Eric89GXL: ...by that I mean you instead would do `ax.plot(...)`

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    plt.title('Mean cHPI power / mean residual variance (gradiometers)')
+    plt.legend(np.array(lines1)[sind], np.array(cfreqs_legend)[sind],
+               prop={'size': legend_fontsize})
+    plt.ylabel('SNR (dB)')
```
Eric89GXL: ...and `ax.set_ylabel(...)` or better `ax.set(ylabel=..., xlabel=..., xlim=..., ...)`

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    plt.xlabel('Time (s)')
+    # create some horizontal space for legend
+    plt.xlim([plt.xlim()[0], plt.xlim()[1]*(1+legend_hspace/100.)])
+
+    plt.figure()
```
Eric89GXL: better to use `fig, axs = plt.subplots(2, 1)` or `fig, axs = plt.subplots(1, 2)` so there is only one window with multiple subplots

mnefun/_mnefun.py
```diff
@@ -2330,3 +2330,117 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    lines1 = plt.plot(tvec, 10*np.log10(snr_avg_mag.transpose()))
+    plt.title('Mean cHPI power / mean residual variance (magnetometers)')
+    plt.legend(np.array(lines1)[sind],
+               np.array(cfreqs_legend)[sind], prop={'size': legend_fontsize})
+    plt.ylabel('SNR (dB)')
+    plt.xlabel('Time (s)')
+    plt.xlim([plt.xlim()[0], plt.xlim()[1]*(1+legend_hspace/100.)])
+
+    # residual (unexplained) variance as function of time
+    plt.figure()
```
Eric89GXL: or I guess `plt.subplots(3, 1)`, etc.

bin/chpi_snr.py
```diff
@@ -0,0 +1,46 @@
...
+    parser.add_argument('--winlen', type=float, default=default_winlen,
+                        help='Buffer length for SNR estimates (s)')
+    parser.add_argument('--nharm', type=int, default=default_nharm,
+                        choices=[0, 1, 2, 3, 4], help='Number of line'
+                        'frequency harmonics to include')
+    parser.add_argument('--stop', type=float, metavar='t', default=None,
+                        help='Process only first t seconds')
+    args = parser.parse_args()
+
+    raw = mne.io.Raw(args.fiff_file, allow_maxshield=True)
```
Eric89GXL: allow_maxshield='yes' to avoid warnings

bin/chpi_snr.py
```diff
@@ -0,0 +1,46 @@
...
+    chpi_snr_raw(raw, args.winlen, args.nharm, args.stop)
```
Eric89GXL: should probably call it `plot_chpi_snr_raw` since it plots

bin/chpi_snr.py
```diff
@@ -0,0 +1,46 @@
...
+"""
+Plot signal-to-noise of continuous HPI coils as a function of time.
+Works by fitting a general linear model (HPI freqs, line freqs, DC, slope) to
+the data, and comparing estimated HPI powers with the residual (=variance
+unexplained by the model).
+Window length for SNR estimates can be specified on the command line.
+Longer windows will by nature include more low frequencies and thus have
+larger residual variance (lower SNR).
+
+Tested with Python 2.7, MNE 0.11.0
```
Eric89GXL: Don't really need this line, it will change over time anyway

0.11 is pretty old now BTW, I recommend using 0.12 or even latest `master` if you're going to be developing

Eric89GXL:
Despite a bunch of little MNE-Python/mnefun standardization stuff LGTM

Eric89GXL:
... can you add an example to `examples/`? It could use the `funloc` data since I *think* that has cHPI turned on.

mnefun/_mnefun.py
jjnurminen: Not 100% sure what was wrong with the docstring, please check the new one.

jjnurminen:
I couldn't find the funloc raw data to test the example script. I guess it's not in the repo. Otherwise everything is ok.

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,114 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+        Length of window to use. Longer windows will naturally include more
+        low frequency power, resulting in lower SNR.
+    nharm : int
+        Number of line frequency harmonics to include in the model.
+    stop : int
+        If specified, stop before the end of file (seconds)
+
+    Returns
+    -------
+    fig : Instance of matplotlib.figure.Figure
```
Eric89GXL: docstring looks better now, other than the capitalization of Instance -> instance (just by convention)

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,114 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
     fig.tight_layout()
     return fig
+
+
+def plot_chpi_snr_raw(raw, winlen, nharm, stop=None):
+    """Compute and plot cHPI SNR from raw data
+
+    Parameters
+    ----------
+    winlen : int
```
Eric89GXL: this should probably be `duration` in seconds, and convert it to samples internally

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,114 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
     axs[2, 0].set_ylabel('Reconstruction')
     fig.tight_layout()
     return fig
+
+
+def plot_chpi_snr_raw(raw, winlen, nharm, stop=None):
```
Eric89GXL: Remove the `stop` argument as one can always do this instead:
```
plot_chpi_snr_raw(raw.copy().crop(0, stop), duration, n_harmonics)
```

Eric89GXL:
Please also add your script to the [setup.py script] (https://github.com/LABSN/mnefun/blob/master/setup.py#L41) so it installs properly for people. Otherwise LGTM, I'll give it a try Monday

mnefun/_mnefun.py
jjnurminen: It is already in seconds (which should be mentioned in the docstring) and gets converted to samples.
I don't like `duration` that much, isn't `winlen` more descriptive? (Maybe it should be `window_len`)

mnefun/_mnefun.py
jjnurminen: Good point.

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,108 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
     fig.tight_layout()
     return fig
+
+
+def plot_chpi_snr_raw(raw, win_length, n_harmonics):
+    """Compute and plot cHPI SNR from raw data
+
+    Parameters
+    ----------
+    win_length : int
```
Eric89GXL: `float` not `int`

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,108 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    pick_mag = pick_types(raw.info, meg='mag', exclude=[])
+    pick_grad = pick_types(raw.info, meg='grad', exclude=[])
+    nchan = len(pick_meg)
+
+    # create general linear model for the data
+    t = np.arange(buflen) / float(sfreq)
+    model = np.c_[t, np.ones(t.shape)]  # model slope and DC
+    # add sine and cosine term for each freq
+    for f in np.concatenate([linefreqs, cfreqs]):
+        model = np.c_[model, np.cos(2*np.pi*f*t), np.sin(2*np.pi*f*t)]
```
Eric89GXL: This is memory inefficient and ambiguous about the resulting size (it needs to be mentally calculated rather than being clear from the code), better to preallocate `model = np.empty((len(t), 2 + len(linefreqs) + len(cfreqs)))` and populate it using indices. I know it probably doesn't make a noticable performance difference here, but it's better to get in good habits early :)

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,108 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    legend_hspace = 25  # % of horizontal axis to reserve for legend
+
+    # get some info from fiff
+    sfreq = raw.info['sfreq']
+    linefreq = raw.info['line_freq']
+    linefreqs = (np.arange(n_harmonics + 1) + 1) * linefreq
+    buflen = int(win_length * sfreq)
+    if buflen <= 0:
+        raise ValueError('Window length should be >0')
+    (cfreqs, _, _, _, _) = _get_hpi_info(raw.info)
```
Eric89GXL: Some of the other outputs here are related to whether or not cHPI was actually active (sometimes they aren't for the first few). If you want this to be more complete and robust, you should probably skip buffers in your `for ind, buf0 in ...` loop below, similar to this:

https://github.com/mne-tools/mne-python/blob/master/mne/chpi.py#L373

Eric89GXL: (but that can also be a separate PR, just make a comment in the code about it as a reminder)

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,108 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    for ind, buf0 in enumerate(bufs):
+        print('Buffer %s/%s' % (ind+1, len(bufs)))
+        megbuf = raw[pick_meg, buf0:buf0+buflen][0].T
+        coeffs = np.dot(inv_model, megbuf)
+        coeffs_hpi = coeffs[2+2*len(linefreqs):]
+        resid_vars[:, ind] = np.var(megbuf-np.dot(model, coeffs), 0)
+        # get total HPI amplitudes by combining sine and cosine terms
+        hpi_amps = np.sqrt(coeffs_hpi[0::2, :]**2 + coeffs_hpi[1::2, :]**2)
+        # divide average HPI power by average variance
+        snr_avg_grad[:, ind] = (hpi_amps**2/2)[:, pick_grad].mean(1) / \
```
Eric89GXL: better to do something like `hpi_powers = hpi_amps * hpi_amps` above rather than recalculating the same quantity here and two lines below

Eric89GXL: ...and why take the `sqrt` above when you just immediately square it here? Better to just do `hpi_powers = (coeffs_hpi...)` above and skip the `sqrt()` and `**2` steps here

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,108 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+            resid_vars[pick_grad, ind].mean()
+        snr_avg_mag[:, ind] = (hpi_amps**2/2)[:, pick_mag].mean(1) / \
+            resid_vars[pick_mag, ind].mean()
```
Eric89GXL: Why divide the power by 2?

mnefun/_mnefun.py
jjnurminen: The code was using amplitudes at some point. Can be fixed now that we're settled on power.

mnefun/_mnefun.py
jjnurminen: Sinusoidal of amplitude A has a RMS power of A**2/2

Eric89GXL:
This is the output I get:
figure_1

A few comments:

1. The residual plot (bottom) could be a bit prettier. I'd suggest using a set of colors like:
    `colors = plt.get_cmap('plasma')(np.linspace(0., 0.7, len(picks)))`, where each line gets one of the colors based on its mean residual value (across time), and apply an alpha of 0.5 or so. That might make it easier to see the trends. It also makes it clearer it has no relation to the colors in the first few plots (which will be given by mpl defaults).
2. Use `ax.set_xlim(...)` to make the bounds tighter
3. Use `from mne.viz import tight_layout ... tight_layout()` to tighten the subplot spacing.
4. Your vertical layout seems to be limited by a large y-axis label on the last plot, I'd do `Var. (...)` instead of `Variance (...)`
4. You could put the `xlabel` on only the last axis to give the plots more vertical space.

Eric89GXL:
...also, instead of `^2` in the ylabel you can do `$^2$` to get an actual superscript, as it will render in LaTeX mode

mnefun/_mnefun.py
Eric89GXL: Ahh right. Can you add a comment so that future reviewers don't embarrass themselves? :)

mnefun/_mnefun.py
jjnurminen: I'll think about it. The fact that cHPI is off should also be obvious from the SNR plot. I'm not that keen on skipping buffers, since the user is probably interested in seeing the whole time range.

Eric89GXL: How about taking the time period where cHPI was off and adding a `fill_between` in light gray in the background? That way a user looking at a random file will immediately know that the terrible cHPI values are due to it being off as opposed to, say, electronic failure?

jjnurminen:
Somehow it looks worse than it looks here. Y scale of SNR is missing and the legends are too large. I'll look into it.

The code already calls `tight_layout()` from Matplotlib, is the MNE version somehow smarter?

mnefun/_mnefun.py
Eric89GXL: (background plotting can be achieved by using e.g. `zorder=1` since lines by default get `zorder=2`)

mnefun/_mnefun.py
jjnurminen: Why not ;)

Eric89GXL:
> Somehow it looks worse than here. Y scale of SNR is missing and the legends are too large.

I am running the latest (2.0-esque) `matplotlib`. You should be able to make the legend smaller by explicitly setting the text size to perhaps 8.

> The code already calls tight_layout() from Matplotlib, is the MNE version somehow smarter?

Yes, we have some workarounds in there.

I actually assumed `tight_layout` wasn't in there from looking at the output (silly me didn't look at the code), maybe it's because the Y-limits are currently determined by the Y-axis label as opposed to the plotting areas.

Eric89GXL:
(I think you can more or less safely ignore the bad `ytick`ing, that's matplotlib's fault and will probably be fixed before 2.0 release)

Eric89GXL:
And for bonus points, I wonder if you could do a legend like one of these outside the `axes` area:

http://stackoverflow.com/questions/4700614/how-to-put-the-legend-out-of-the-plot

Otherwise I fear there will be overlap between the lines and the legend.

mnefun/_mnefun.py
jjnurminen: Personally I'd prealloc only when needed and skip it here to keep the code more concise. But the customer is always right ;)

mnefun/_mnefun.py
jjnurminen: It's a good idea, bu right now t I don't have time to figure out how to parse the cHPI on/off status. If it's ok with you, I'll make it into a separate PR later.

jjnurminen:
chpi_snr_plot

That's how it looks now at my end. Matplotlib 1.5.something. Oops, the titles are smaller than ylabels now.

jjnurminen:
chpi_snr_plot

This is how it looks now at my end. Matplotlib 1.5.something.

bin/plot_chpi_snr.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Plot signal-to-noise of continuous HPI coils as a function of time.
+Works by fitting a general linear model (HPI freqs, line freqs, DC, slope) to
+the data, and comparing estimated HPI powers with the residual (=variance
+unexplained by the model).
+Window length for SNR estimates can be specified on the command line.
```
Eric89GXL: empty newline between paragraphs otherwise it gets rendered as a single paragraph

bin/plot_chpi_snr.py
```diff
@@ -0,0 +1,41 @@
...
+Longer windows will by nature include more low frequencies and thus have
+larger residual variance (lower SNR).
+"""
+
+import sys
+import os
+sys.path.append(os.path.join(os.path.dirname(__file__), os.pardir))
```
Eric89GXL: You shouldn't need to do this, why do you need it?

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,144 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+def plot_chpi_snr_raw(raw, win_length, n_harmonics):
+    """Compute and plot cHPI SNR from raw data
+
+    Parameters
+    ----------
+    win_length : float
+        Length of window to use for SNR estimates (seconds). A longer window
+        will naturally include more low frequency power, resulting in lower
+        SNR.
+    n_harmonics : int
```
Eric89GXL: If you want to go the extra mile this could be "auto" to use as many up to Nyquist freq or `info['lowpass'`], whichever is lower

Eric89GXL: (or "extra kilometer" I guess if you want to be metric about it)

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,144 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+        Number of line frequency harmonics to include in the model.
+
+    Returns
+    -------
+    fig : Instance of matplotlib.figure.Figure
```
Eric89GXL: lower case "I" on "instance"

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,144 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+        cHPI SNR as function of time, residual variance.
+
+    Notes
+    -----
+    A general linear model including cHPI and line frequencies is fit into
+    each data window. The cHPI power obtained from the model is then divided
+    by the residual variance (variance of signal unexplained by the model) to
+    obtain the SNR.
+    The SNR may decrease either due to decrease of cHPI amplitudes (e.g.
```
Eric89GXL: Same thing about empty newline between paragraphs

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,144 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+
+    # get some info from fiff
+    sfreq = raw.info['sfreq']
+    linefreq = raw.info['line_freq']
+    linefreqs = (np.arange(n_harmonics + 1) + 1) * linefreq
+    buflen = int(win_length * sfreq)
+    if buflen <= 0:
+        raise ValueError('Window length should be >0')
+    (cfreqs, _, _, _, _) = _get_hpi_info(raw.info)
+    print('\nNominal cHPI frequencies: %s Hz' % cfreqs)
```
Eric89GXL: Nothing has been printed before this so it's weird to start with a newline

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,144 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    model = np.empty((len(t), 2+2*(len(linefreqs)+len(cfreqs))))
+    model[:, 0] = t
+    model[:, 1] = np.ones(t.shape)
+    # add sine and cosine term for each freq
+    allfreqs = np.concatenate([linefreqs, cfreqs])
+    ind = 2
+    for f in allfreqs:
+        model[:, ind] = np.cos(2*np.pi*f*t)
+        model[:, ind+1] = np.sin(2*np.pi*f*t)
+        ind += 2
```
Eric89GXL: to save some lines (no need for `ind`):
```
for fi, f in enumerate(allfreqs):
    model[:, 2 * fi + 2] = ...
    model[:, 2 * fi + 3] = ...
```
Or if you want to be fancy use [broadcasting] (http://docs.scipy.org/doc/numpy/user/basics.broadcasting.html) to avoid the loop entirely:
```
model[:, 2::2] = np.cos(2 * np.pi * f[:, np.newaxis] * t)
model[:, 3::2] = np.sin(2 * np.pi * f[:, np.newaxis] * t)
```

mnefun/_mnefun.py
```diff
@@ -2330,3 +2332,144 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+        print('Buffer %s/%s' % (ind+1, len(bufs)))
+        megbuf = raw[pick_meg, buf0:buf0+buflen][0].T
+        coeffs = np.dot(inv_model, megbuf)
+        coeffs_hpi = coeffs[2+2*len(linefreqs):]
+        resid_vars[:, ind] = np.var(megbuf-np.dot(model, coeffs), 0)
+        # get total power by combining sine and cosine terms
+        # sinusoidal of amplitude A has power of A**2/2
+        hpi_pow = (coeffs_hpi[0::2, :]**2 + coeffs_hpi[1::2, :]**2)/2
+        # divide average HPI power by average variance
+        snr_avg_grad[:, ind] = (hpi_pow)[:, pick_grad].mean(1) / \
```
Eric89GXL: no need for parens on `hpi_pow`

bin/plot_chpi_snr.py
Eric89GXL: (now that you've added your script to `setup.py`, you might need to re-run `python setup.py develop` to get it to install a wrapper to your script)

Eric89GXL:
Looking at that plot, I think 4 rows would actually be useful:

1. Ratio grads
2. Ratio mags
3. cHPI grads (new one)
4. Residual grads (currently 3)

That way it's clear which dips are due to the residual growing, and which are due to the cHPI coefficients shrinking.

Otherwise LGTM

Eric89GXL:
Thanks for sticking with it even though I'm making many requests for changes. Assuming you eventually want to get code into MNE-Python you'll likely see a similar barrage, but the barrage generally gets smaller and smaller as the conventions sink in. Then hopefully you'll experience a bit of Stockholm syndrome and learn to love the conventions :)

mnefun/_mnefun.py
jjnurminen: Extra 1.609 km?
I'm not 100% sure this makes sense. With 5 kHz it would be dozens of harmonics, majority of which are probably negligible. I can look into it though.

Eric89GXL: What would be the most reasonable default for "auto"? Something ilke all harmonics up to 400 Hz (or lower depending on `info['lowpass'`] and `info['sfreq']`? That would probably get most of the bad ones in most of the cases, right?

jjnurminen:
> Looking at that plot, I think 4 rows would actually be useful

Head movements always result in broadband interference (muscle activation, movement artifacts), which probably is the main cause of the SNR decrease. The coil amplitudes may also decrease, but I think that's a secondary effect in most cases. Anyway, it might be nice to plot the amplitudes also.

> Thanks for sticking with it even though I'm making many requests for changes.

I don't mind, it's also a learning experience for me. Unfortunately I cannot devote as much resources to this as I would like. But I firmly believe it'll be mergeable some day ;)

mnefun/_mnefun.py
jjnurminen: Yep, for example. I could look at some 5 kHz data we just recorded. IIRC the line freq harmonics don't typically go very high, even though that may vary by country / electricity provider...

Eric89GXL: That's fine, it doesn't have to be perfect since people can always override it

bin/plot_chpi_snr.py
jjnurminen: The idea was to be able to run the script either from bin/ or the parent dir. If `setup.py` takes care of this, I guess the path append is not needed.

mnefun/_mnefun.py
jjnurminen: BTW if I understand it correctly, info['lowpass'] is supposed to be the corner frequency (-3 dB?) of the analog anti-aliasing filter, and thus it should always be clearly below Nyquist frequency. With the Elekta system, you cannot set the lowpass corner above `fs/3` (where `fs` is sampling frequency). OTOH, if there is strong interference at frequencies above the lowpass corner, the analog filter may not be able to get rid of it.

I'll see what happens when including freqs upto `info['lowpass'].` or even Nyquist by default. If it works fine, why not.

mnefun/_mnefun.py
jjnurminen: broadcasting is always cool, of course...

jjnurminen:
Linefreq harmonics still wip, everything else should be there.

Eric89GXL:
We can make that a separate PR for later if you want. I suspect you have more interesting things to work on :)

Eric89GXL:
It's a little bit cramped by default, can you try reducing title and ylabel sizes? You could try doing `$\mu_{cHPI} / \mu_{res}` or something for the first two titles, and maybe try newlines to separate label from unit in the ylabels...?

figure_1

You could also manually set the `figsize=(w, h)` to be a bit taller

jjnurminen:
Looks disgusting. I added some mysterious pad options to `tight_layout` and I get this:

figure_7

How does it look for you?

jjnurminen:
Actually the onscreen version looked better, the png rendering is slightly more crowded. Onscreen it looks pretty good at my end.

Eric89GXL:
Good enough for me:

figure_1

jjnurminen:
Your matplotlib seems to set title font sizes bigger than me. Maybe I can set both title and label fontsizes explicitly.

Eric89GXL:
Yep, you can do that, but it will still vary based on DPI and the like. A better bet is to make the titles shorter

jjnurminen:
figure_1

I like it now, except that matplotlib does something clever with the 1e-20 multiplier and its fontsize does not change. Sigh :)

jjnurminen:
Can also shorten the titles, but I think easily understandable titles are usually best.

Eric89GXL:
figure_1

The x-axes are not tight, though, so you should probably `ax.set(xlim=...)` in there.

BTW since all x axes are shared you could link them all, that way zooming X in one zooms them all.

jjnurminen:
Not sure how you can live with such a weird matplotlib ;)

jjnurminen:
`sharex` is cool, I wasn't aware of that.

Eric89GXL:
Living on the edge like this is indeed a sacrifice (it is unbearable for some), but I feel it is my duty to ensure the future safety of our software designs

figure_1

I'm happy to merge if you're happy with it

jjnurminen:
Fine with me. Thanks for helping to improve it a lot.

Eric89GXL:
Thanks for sticking with it @jjnurminen

PR #136, opened 2016-07-04
use harmonics up to lowpass by default

jjnurminen:
WIth my test data, results are almost identical with using just a couple of harmonics. Still it's probably a good default.


Eric89GXL:
LGTM, thanks @jjnurminen

Issue #140, opened 2016-08-08
Projection vector "Average EEG reference"


kambysese:
@Eric89GXL fu on 

> RuntimeWarning: Projection vector "Average EEG reference" has magnitude 0.12 (should be unity), applying projector with 53/60 of the original channels available may be dangerous, consider recomputing and adding projection vectors for channels that are eventually >used. If this is intentional, consider using info.normalize_proj() raw.apply_proj()

I still get this warning when `pre_bad.txt` & `post-sss.txt` files are identical. On a related note, is it the expected behavior that channels in the `pre_bad` end up in the `post-sss` file, i.e., are the pre-bad channels not excluded from the outset of processing?

Eric89GXL:
The pre-bad MEG channels should *not* be marked bad in the SSS'ed file. SSS should reconstruct them.

We must be adding the avg eeg ref at the wrong time. We probably shouldn't add it on read, but rather read, add bads to info, then add avg eeg ref.

kambysese:
We meaning mne-python or mnefun end? FYI I am hitting this on 2 of 29 datasets, so not sure if it's a corner case. 
Not understanding this. Why should SSS reconstruct bad channels? I was under the impression tsss operation spread bad channel data across sensors. Are you also implying that it is a bad idea to put channels (i.e., noisy channels that cause numerous trial drops) in the prebad list?

Eric89GXL:
> Why should SSS reconstruct bad channels?

Any channels in `raw.info['bads']` when calling `maxwell_filter(raw, ...)` will simply be excluded during the multipole estimation step, but are easily reconstructed (and removed from `raw_sss.info['bads']`) from the multipole moments during the reconstruction step.

> I was under the impression tsss operation spread bad channel data across sensors

Yes, so you should mark any MEG channels that you know are pure noise or have jumpy artifacts in `raw_fif/subj_prebad.txt`. That is actually the only reason to have the prebad file.

> Are you also implying that it is a bad idea to put channels (i.e., noisy channels that cause numerous trial drops) in the prebad list?

There is no reason to put EEG data in the prebad list, for example. The only thing that prebad should be used for is excluding MEG channels from multipole moment estimation during SSS.

Eric89GXL:
> We meaning mne-python or mnefun end?

mnefun

kambysese:
I can't find where...

[...]

PR #141, opened 2016-08-09
FIX autobad channel detection without eeg average reference projection

kambysese:
Should do it #140

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1660,7 +1660,7 @@ def _raw_LRFCP(raw_names, sfreq, l_freq, h_freq, n_jobs, n_jobs_resample,
                projs, bad_file, disp_files=False, method='fft',
                filter_length=32768, apply_proj=True, preload=True,
                force_bads=False, l_trans=0.5, h_trans=0.5,
-               allow_maxshield=False, eeg_ref=True):
+               allow_maxshield=False):
     """Helper to load, filter, concatenate, then project raw files
     """
     if isinstance(raw_names, str):
@@ -1670,9 +1670,10 @@ def _raw_LRFCP(raw_names, sfreq, l_freq, h_freq, n_jobs, n_jobs_resample,
     raw = list()
     for rn in raw_names:
         r = Raw(rn, preload=True, allow_maxshield='yes',
-                add_eeg_ref=eeg_ref)
-        r.pick_types(meg=True, eeg=True, eog=True, ecg=True, exclude=())
+                add_eeg_ref=False)
         r.load_bad_channels(bad_file, force=force_bads)
+        r.pick_types(meg=True, eeg=True, eog=True, ecg=True, exclude=[])
+        r.add_eeg_average_proj()
         if sfreq is not None:
             r.resample(sfreq, n_jobs=n_jobs_resample, npad='auto')
         if l_freq is not None or h_freq is not None:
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -1730,12 +1731,13 @@ def do_preprocessing_combined(p, subjects, run_indices):
                   '        %s' % bad_file)
             if not op.isdir(bad_dir):
                 os.mkdir(bad_dir)
-            # do autobad
+            # do autobad without applying eeg average ref projection
```
Eric89GXL: Hmm... we probably always want to apply it, just do it at the right time (after setting bads instead of before)

kambysese: So auto bad ch selection should happen with EEG avg ref proj applied?

Eric89GXL: I think so, yeah

kambysese: But then we're back in the situation where the avg ref proj may include eeg chs that are declared bad, no?

Eric89GXL: Read with add=False, mark bad, then add projector. When it gets added at
that stage it shouldn't include the bads

kambysese: Ok. Is there a way in mne python to delay application of the projector (specifically the eeg avg ref)? otherwise it sounds like we might need a helper file to do this.

mnefun/_mnefun.py
```diff
@@ -1818,7 +1817,11 @@ def do_preprocessing_combined(p, subjects, run_indices):
             n_jobs=p.n_jobs_fir, n_jobs_resample=p.n_jobs_resample,
             projs=projs, bad_file=bad_file, disp_files=p.disp_files,
             method='fft', filter_length=p.filter_length, force_bads=False,
-            l_trans=p.hp_trans, h_trans=p.lp_trans)
+            l_trans=p.hp_trans, h_trans=p.lp_trans, eeg_ref=False)
+        for ii in raw_orig.info['projs']:
+            if ii['desc'] == 'Average EEG reference':
+                eeg_ref_proj = ii
+                raw_orig.add_proj(eeg_ref_proj).apply_proj()
```
kambysese: This also works with my dataset, did not test with example data b/c I am not sure if 
`for ii in raw_orig.info['projs']:
             if ii['desc'] == 'Average EEG reference':
                 eeg_ref_proj = ii
                 raw_orig.add_proj(eeg_ref_proj).apply_proj()`

is necessary; seems like `raw_orig.apply_proj()` should suffice after the read?

Eric89GXL: I would probably do:
```
raw.remove_proj(ii)
http://mne-tools.github.io/stable/generated/mne.io.Raw.html#mne.io.Raw.add_eeg_average_proj
```
and `proj` is a better name for the thing to iterate instead of `ii`, which is usually used for an index

But really this probably isn't the right place for this, but rather where we set the bad channels from the `bads/*.txt` file

kambysese: To make sure we're on the same page, doesn't the `_raw_LRFCP` call on ln 1815 set the bad chs as per `bad_file` without the eeg_average_proj applied?

Eric89GXL: Yes, but it's not about when `apply_proj` is called, it's about when the projector is created and added to the measurement info.

Eric89GXL: ... so this should maybe go in `_raw_LRFCP`

kambysese:
So I am pretty sure it was just the case where autobad selection was not executed correctly. In addition to the minor changes in `_raw_LRFCP`, calling it `_raw_LRFCP` with `apply_proj=True` resolves the RT warning.   I stepped through the code and verified that now raw is read without eeg avg-ref proj, bad chs are added to measurement info and the proj is added and applied subsequently. LMKWYT

Eric89GXL:
Thanks @kambysese

PR #142, opened 2016-08-12
[ENH] Save_epochs fail gracefully if no epochs left.

### LINKED ISSUE #139: save_epochs raises error if all events rejected
https://github.com/LABSN/mnefun/blob/master/mnefun/_mnefun.py#L1353

```
offset = max(epochs.events[:, 2].max(), new_numbers.max()) + 1
```

This line in `save_epochs` causes an error;

`zero-size array to reduction operation maximum which has no identity`;

 if the `epochs.events` array is empty, e.g. when all epochs were rejected.

ilogue:
fixes #139

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,5 @@ examples/funloc/run_sss.sh
 
 *~
 /examples/funloc/
+env/
+*.whl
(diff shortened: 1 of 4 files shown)


Eric89GXL:
Sorry I was looking at an old commit

Eric89GXL:
Thanks @ilogue

PR #146, opened 2016-09-30
Add test data

kambysese:
I can't figure what the fuck happened to this PR


Eric89GXL:
What are you trying to do with this PR? Test `mne_process_raw` it looks like?

kambysese:
IIRC you were interested in writing tests for mnefun at some point and I thought that would be a good way for me to learn how to write tests b/c I seem to be completely incapable of conceiving of test routines. So at some point we decided it may be useful to have some `testdata` available for such tests. Obviously I've been unable to follow up with this PR, I guess at this point we can revisit the issue and determine if testing data may in fact be useful; and if so decide on specifics. I thought it would be a good idea to emulate for the most part what is available in mne-python. LMKWYT & and what is `mne_process_raw`?

Eric89GXL:
> what is mne_process_raw

It is in the `mnefun/tests/data/process_raw.sh` from this PR, not sure why. It's the MNE-C raw data processing program.

Since this has a lot of cruft apparently, and a merge commit in the middle, let's close it and start clean. Let's add data once we have some tests to write. Did you have some in mind?

kambysese:
Well, Myles' QC script is on the docket and I am thinking of adding it to the MNEFUN pipeline after epoching, WDYT? That might need a test, but I am not sure what the exact routine is yet.
FYI, I uploaded most of the `test-xxx.fif` files from this PR to https://github.com/ilogue/niprov/pull/170 in order to smooth out wrinkles in the provenance tracking.

Eric89GXL:
What does the QC script do?

kambysese:
Not sure exactly. It's the routines Myles runs after acquisition which generate the cHPI SNR plots.

PR #149, opened 2016-10-14
Try fetching data from multiple mounting points on sinuhe acq machine

### LINKED ISSUE #148: Acquisition directories
Provide list of acq directories to `fetch_raw` func.

kambysese:
Fixes #148

### DIFF
--- a/examples/funloc/analysis_fun.py
+++ b/examples/funloc/analysis_fun.py
@@ -47,12 +47,17 @@ params.subjects = ['subj_01', 'subj_02']
 params.structurals = [None, 'AKCLEE_110_slim']  # None means use sphere
 params.dates = [(2014, 2, 14), None]  # Use "None" to more fully anonymize
 params.score = score  # Scoring function used to slice data into trials
-params.subject_indices = np.arange(2)  # Define which subjects to run
+params.subject_indices = [0]  # Define which subjects to run
 params.plot_drop_logs = False  # Turn off so plots do not halt processing
 
 # Set parameters for remotely connecting to acquisition computer
 params.acq_ssh = 'minea'  # Could also be e.g., "eric@minea.ilabs.uw.edu"
-params.acq_dir = 'eric_non_space' # name of remote dir containing data
+# Pass list of paths to search and fetch raw data
+params.acq_dir = ['/sinuhe_data01/eric_non_space',
+                  '/data101/eric_non_space',
+                  '/sinuhe/data01/eric_non_space',
+                  '/sinuhe/data02/eric_non_space',
+                  '/sinuhe/data03/eric_non_space']
 
 # Set parameters for remotely connecting to SSS workstation ('sws')
 params.sws_ssh = 'kasga'
--- a/examples/funloc/analysis_fun.py
+++ b/examples/funloc/analysis_fun.py
@@ -47,7 +47,7 @@ params.subjects = ['subj_01', 'subj_02']
 params.structurals = [None, 'AKCLEE_110_slim']  # None means use sphere
 params.dates = [(2014, 2, 14), None]  # Use "None" to more fully anonymize
 params.score = score  # Scoring function used to slice data into trials
-params.subject_indices = [0]  # Define which subjects to run
+params.subject_indices = np.arange(2)  # Define which subjects to run
 params.plot_drop_logs = False  # Turn off so plots do not halt processing
 
 # Set parameters for remotely connecting to acquisition computer
--- a/examples/funloc/analysis_fun.py
+++ b/examples/funloc/analysis_fun.py
@@ -52,7 +52,7 @@ params.plot_drop_logs = False  # Turn off so plots do not halt processing
 
 # Set parameters for remotely connecting to acquisition computer
 params.acq_ssh = 'minea'  # Could also be e.g., "eric@minea.ilabs.uw.edu"
-params.acq_dir = '/net/sinuhe/data02/eric_non_space'
+params.acq_dir = 'eric_non_space' # name of remote dir containing data
 
 # Set parameters for remotely connecting to SSS workstation ('sws')
 params.sws_ssh = 'kasga'
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -567,6 +567,13 @@ def do_processing(p, fetch_raw=False, do_score=False, push_raw=False,
     assert all(r is None or np.in1d(r, np.arange(len(p.run_names))).all()
                for r in run_indices)
 
+    if isinstance(p.acq_dir, string_types):
+        p.acq_dir = [loc + '/%s' % p.acq_dir for loc in ['/sinuhe_data01',
+                                                         '/data101',
+                                                         '/sinuhe/data01',
+                                                         '/sinuhe/data02',
+                                                         '/sinuhe/data03']]
+
     # Actually do the work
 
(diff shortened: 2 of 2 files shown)


kambysese:
@Eric89GXL ran into naming issues with `fname[len(p.acq_dir):].lstrip('/')` indexing during local file management after fetch. Newest commit works for my own datasets and funloc example.

mnefun/_mnefun.py
```diff
@@ -610,18 +610,24 @@ def fetch_raw_files(p, subjects, run_indices):
...
-        finder_stem = 'find %s ' % p.acq_dir
-        # build remote raw file finder
-        fnames = get_raw_fnames(p, subj, 'raw', True, False, run_indices[si])
+        fnames = get_raw_fnames(p, subj, 'raw', True, False,
+                                run_indices[si])
         assert len(fnames) > 0
-        finder = (finder_stem +
+        # build remote raw file finder
+        for remote_dir in ['/sinuhe_data01', '/data101', '/sinuhe/data01',
+                           '/sinuhe/data02','/sinuhe/data03']:
```
kambysese: Ok.
I am having trouble remembering how to modify values of `params` attributes.
Line 378 in `__init__` is `self.acq_ssh = self.acq_dir = None` which is overridden by users arg I am failing to see why adding:
```
if acq_dir is not None:
self.acq_dir = [loc+'/%s' % self.acq_dir for loc in ['/sinuhe_data01',
                                                                             '/data101',
                                                                             '/sinuhe/data01',
                                                                             '/sinuhe/data02',
                                                                             '/sinuhe/data03']]
```
on the next line fails to change the value for `acq_dir`?

Eric89GXL: Where are you putting that conditional? Maybe it would help if you pushed code with updated example that you think should work, but errors out

kambysese:
I didn't manage to figure out how to modify `acq_dir` inside `__init__` code for `params`. Recent commit still hard codes list inside `do_preprocessing` setup; and there is still the problem of catching errors when `acq_dir` doesn't exist on one the resulting paths e.g., /sinuhe_data01/eric_non_space. I can handle that when you're happy with where the list actually ends up.

examples/funloc/analysis_fun.py
```diff
@@ -52,7 +52,7 @@
 
 # Set parameters for remotely connecting to acquisition computer
 params.acq_ssh = 'minea'  # Could also be e.g., "eric@minea.ilabs.uw.edu"
-params.acq_dir = '/net/sinuhe/data02/eric_non_space'
+params.acq_dir = 'eric_non_space' # name of remote dir containing data
```
Eric89GXL: So the change would be that `params.acq_dir` can now be a `str` or `list of str`

kambysese: No comment :confounded:

kambysese:
Please review, think it's done.

mnefun/_mnefun.py
```diff
@@ -610,20 +610,26 @@ def fetch_raw_files(p, subjects, run_indices):
...
-        finder_stem = 'find %s ' % p.acq_dir
-        # build remote raw file finder
-        fnames = get_raw_fnames(p, subj, 'raw', True, False, run_indices[si])
+        fnames = get_raw_fnames(p, subj, 'raw', True, False,
+                                run_indices[si])
         assert len(fnames) > 0
-        finder = (finder_stem +
+        # build remote raw file finder
+        assert type(p.acq_dir) in (list, string_types)
+        for remote_dir in p.acq_dir:
```
Eric89GXL: this will no longer work for `string_types`. You probably want:
```
use_dirs = [p.acq_dir] if isinstance(p.acq_dir, string_types) else p.acq_dir
for remote_dir in use_dirs:
    ...
```

mnefun/_mnefun.py
```diff
@@ -610,20 +610,26 @@ def fetch_raw_files(p, subjects, run_indices):
...
+            finder_stem = 'find %s ' % remote_dir
+            finder = (finder_stem +
                   ' -o '.join(['-type f -regex ' + _regex_convert(f)
                                for f in fnames]))
-        stdout_ = run_subprocess(['ssh', '-p', str(p.acq_port),
-                                  p.acq_ssh, finder])[0]
+            stdout_ = run_subprocess(['ssh', '-p', str(p.acq_port),
+                                      p.acq_ssh, finder])[0]
+            if len(stdout_) > 0:
+                break
```
Eric89GXL: This `break` method will only work if the files are only in one location, not if they are spread across multiple locations. This can be fixed with something like:
```
remote_fnames = list()
for remote_dir in use_dirs:
    ...
    these_fnames = [x.strip() for x in stdout_.splitlines()]
    assert all(fname.startswith(remote_dir + '/') for fname in these_fnames)
    remote_fnames += these_fnames
```
But really this is even a bit clunky, since you can [make find search multiple directories at the same time] (http://alvinalexander.com/blog/post/linux-unix/find-how-search-multiple-folders-directories-unix). That should be the cleanest solution, and will probably require the fewest changed lines (at least here).

Eric89GXL:
FYI I just split the data so that `subj_02` is on sinuhe's `/data02` and `subj01` is on sinuhe's `/data03` so that you can test that having multiple subjects across different directories works properly

kambysese:
LMKWYT fetching succeeded for the split `funloc` data test case with single `find` and `rsync` commands. Not exactly sure with what you meant by  

> use_dirs = [p.acq_dir] if isinstance(p.acq_dir, string_types) else p.acq_dir
> for remote_dir in use_dirs:

But did something like it...I think.

mnefun/_mnefun.py
```diff
@@ -610,21 +610,27 @@ def fetch_raw_files(p, subjects, run_indices):
...
+                                run_indices[si])
         assert len(fnames) > 0
-        finder = (finder_stem +
-                  ' -o '.join(['-type f -regex ' + _regex_convert(f)
-                               for f in fnames]))
+        # build remote raw file finder
+        if isinstance(p.acq_dir, string_types):
+            use_dir = [p.acq_dir]
+        else:
+            use_dir = p.acq_dir
```
Eric89GXL: yes exactly, the idea is just to ensure it's a list `list` to simplify subsequent operations

mnefun/_mnefun.py
```diff
@@ -610,21 +610,27 @@ def fetch_raw_files(p, subjects, run_indices):
...
+        finder = (finder_stem + ' -o '.join([' -type f -regex ' +
+                                             _regex_convert(f)
+                                             for f in fnames]))
         stdout_ = run_subprocess(['ssh', '-p', str(p.acq_port),
                                   p.acq_ssh, finder])[0]
         remote_fnames = [x.strip() for x in stdout_.splitlines()]
-        assert all(fname.startswith(p.acq_dir) for fname in remote_fnames)
-        # make the name "local" to the acq_dir, so that the name works
+        #assert all(fname.startswith(remote_dir + '/')
+        #           for fname in remote_fnames)
```
Eric89GXL: can you put this back but make it actually work? something like
```
assert all(any(fname.startswith(rd + '/') for rd in use_dir)
           for fname in remote_fnames)
```

Eric89GXL:
I'll give this a try tomorrow and merge if it works

Eric89GXL:
Works great, thanks @kambysese

PR #151, opened 2016-10-24
ENH: plot chpi_snr medians

kambysese:
@Eric89GXL @staulu asked to median traces in plots generated by `plot_chpi_snr_raw`. If you're ok with this, could you give me a clue as to add `--- Median` to the two legends in question.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1042,7 +1042,8 @@ def run_sss_locally(p, subjects, run_indices):
                 ext_order=p.ext_order, calibration=cal_file,
                 cross_talk=ct_file, st_correlation=p.st_correlation,
                 st_duration=st_duration, destination=trans_to,
-                coord_frame='head', head_pos=pos, regularize=reg)
+                coord_frame='head', head_pos=pos, regularize=reg,
+                bad_condition='warning')
             print('%i sec' % (time.time() - t0,))
             raw_sss.save(o, overwrite=True, buffer_size_sec=None)
         #  process erm files if any
@@ -2495,8 +2496,7 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics=None):
     # SNR plots for gradiometers and magnetometers
     ax = axs[0]
     lines1 = ax.plot(tvec, 10*np.log10(snr_avg_grad.transpose()))
-    lines1_med = ax.plot(tvec, 10*np.log10(np.median(snr_avg_grad,
-                                               axis=0).transpose()),
+    lines1_med = ax.plot(tvec, 10*np.log10(np.median(snr_avg_grad, axis=0)),
                          lw=3, ls='--', color='k', label='Median')
     ax.set_xlim([tvec.min(), tvec.max()])
     ax.set(ylabel='SNR (dB)')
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -2495,6 +2495,9 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics=None):
     # SNR plots for gradiometers and magnetometers
     ax = axs[0]
     lines1 = ax.plot(tvec, 10*np.log10(snr_avg_grad.transpose()))
+    lines1_med = ax.plot(tvec, 10*np.log10(np.median(snr_avg_grad,
+                                               axis=0).transpose()),
```
Eric89GXL: Do you still need the transpose? It is presumably 1D following the median operation. Also, you can use `.T` instead of `.transpose()`

Eric89GXL:
Have you looked at the [matplotlib legend tutorial] (http://matplotlib.org/users/legend_guide.html)?

kambysese:
> Have you looked at the matplotlib legend tutorial?
I did and from what I could understand `legend` with corresponding iterables artists & labels is how to control legend content however when I do something e.g., 
```
ax.legend([np.array(lines1)[sind], lines1_med], 
              [np.array(cfreqs_legend)[sind], lines1_med],
              prop={'size': legend_fontsize}, bbox_to_anchor=(1.02, 0.5, ),
              loc='center left')
```
I get _ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()_ and I don't understand what is going on.

Eric89GXL:
`[np.array(lines1)[sind], lines1_med]` is an *iterable of iterables* (list of array), you need it to be an *iterable of handles* (e.g., array of handles). You probably want something like `np.concatenate([np.arary(lines1)[sind], lines1_med])` as the first arg, for example

kambysese:
The way the 2d lines data/legend keys are set up as arrays of integers is making this a real pain in the rear. Your suggestion almost works, and there might be a way to do it using a dict for the handles but it's taking up way too much time.

Eric89GXL:
> The way the 2d lines data/legend keys are set up as arrays of integers

I'm not sure what you mean, `matplotlib` expects lists of handles, then lists of strings, so that's the format you should use. Do you want me to have a look?

kambysese:
Probably not DRY but LGTM @Eric89GXL

Eric89GXL:
Can you make an example script that does this for an `mnefun` subject? That will facilitate testing and use

kambysese:
> Can you make an example script that does this for an mnefun subject?

`chpi_snr_command_line.py` already exists in `examples` which generates the plot for example data. Are you asking for something different?

Eric89GXL:
Ahh right

Eric89GXL:
@staulu is this what you have in mind?

figure_1

staulu:
Sorry about the delay. Yes, that kind of a median curve is useful. I think that I also mentioned to  Kam that it would be illustrative to see the cHPI-SNRs for individual epochs but on second thought Jussi's trial-related weights do the trick.

kambysese:
@staulu the trial-related weights will be made available and come later in the pipeline, whereas these plots will be produced earlier on during processing when users are just beginning to evaluate data quality. @Eric89GXL please merge if you're ok with it.

Eric89GXL:
Thanks @kambysese

PR #152, opened 2016-10-26
fix eeg reference

kambysese:
@Eric89GXL So it looks like `set_eeg_reference` cannot figure out whether EEG channels are in the data or not. You ok with this fix?


kambysese:
Looks like you already addressed this?

Eric89GXL:
I think I did push a fox. Probably should have made a PR, sorry about that

PR #154, opened 2016-11-04
WIP: cHPI weighted averager

jjnurminen:
Integration into mnefun.

### DIFF
--- a/bin/chpi_weighted_average.py
+++ b/bin/chpi_weighted_average.py
@@ -4,18 +4,23 @@
 Average epochs weighted by the signal-to-noise ratio of continuous HPI.
 cHPI SNR is an indicator of the distance between the head and sensors, and
 therefore is correlated with the signal-to-noise of cortical sources.
-
 """
 
+from __future__ import print_function
+
+import os.path as op
+import matplotlib.pyplot as plt
 import mne
 import argparse
 from mnefun import chpi_weighted_average
+from numpy import log10
+import sys
 
 
 def run():
 
     """ Default parameters """
-    default_nharm = 2  # number of line harmonics to include
+    default_nharm = 5  # number of line harmonics to include
 
     """ Parse command line """
     parser = argparse.ArgumentParser(description='Weighted averaging of '
@@ -26,20 +31,23 @@ def run():
                      'the same file, unless --epochs_file is specified.')
     help_epochs_file = ('Raw fiff file to get the epochs from. It must '
                         'have the same epochs/categories as snr_file.')
-    help_reject = ('Whether to use the rejection parameters defined in '
+    help_reject = ('Whether to use the rejection criteria defined in '
                    'the data acquisition when averaging epochs.')
+    help_flat = ('Whether to use the flatness criteria defined in '
+                 'the data acquisition when averaging epochs.')
     help_nharm = 'Number of line frequency harmonics to include.'
     help_epoch_start = 'Epoch start relative to trigger (s)'
     help_epoch_end = 'Epoch end relative to trigger (s)'
     help_plot_snr = 'Whether to plot SNR or not'
     help_stim_channel = 'Which stimulus channel to scan for events'
-    help_sti_mask = 'Mask to apply to the stim channel'
-    parser.add_argument('snr_file', help=help_snr_file)
-    parser.add_argument('--epochs_file', type=str, default=None,
+    help_mask = 'Mask to apply to the stim channel'
+    parser.add_argument('raw_snr', help=help_snr_file)
+    parser.add_argument('--raw_epochs', type=str, default=None,
                         help=help_epochs_file)
     parser.add_argument('--reject', help=help_reject, action='store_true')
+    parser.add_argument('--flat', help=help_flat, action='store_true')
     parser.add_argument('--nharm', type=int, default=default_nharm,
-                        choices=[0, 1, 2, 3, 4], help=help_nharm)
+                        help=help_nharm)
     parser.add_argument('--epoch_start', type=float, default=None,
                         help=help_epoch_start)
     parser.add_argument('--epoch_end', type=float, default=None,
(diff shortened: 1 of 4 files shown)


jjnurminen:
I quickly checked that it runs, but some more testing is needed.
Being able to freely specify rejection and flatness settings would also be good.

mnefun/_mnefun.py
```diff
@@ -2391,6 +2391,194 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    snrs : list of np.array
+        Per-epoch SNR vectors for each category.
+    """
+
+    ap = AcqParserFIF(raw_snr.info)  # should be identical for both files
+
+    evokeds = []
+    snrs = []
+    for cat in ap.categories:
+        print('\nProcessing category: %s' % cat['comment'])
```
kambysese: I think we need to control this by e.g., `if params.in_names is None` set cond using `AcqParserFIF` otherwise create new events as per `params`.

mnefun/_mnefun.py
```diff
@@ -2391,6 +2391,194 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+        if epoch_start is None:
+            epoch_start = cond['tmin']
+        if epoch_end is None:
+            epoch_end = cond['tmax']
+        cond.pop('tmin')  # to avoid duplicate kwargs below
+        cond.pop('tmax')
+
+        reject = ap.reject if reject is True else None
+        flat = ap.flat if flat is True else None
+
```
kambysese: Also, here reject and flat should default to `params.reject` and `params.flat`

mnefun/_mnefun.py
```diff
@@ -2391,6 +2391,194 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+                          nharm=None, epoch_start=None, epoch_end=None,
+                          stim_channel=None, mask=None):
+    """ Average epochs weighted by SNR of continuous HPI (cHPI).
+
+    Parameters
+    ----------
+
+    raw_chpi : instance of Raw
+        The raw instance to get the cHPI SNR from. This file must not be
+        MaxFiltered, because MaxFilter removes the cHPI signals.
```
kambysese: This is not always true i.e., `maxwell_filter` does not remove the cHPI signals.

mnefun/_mnefun.py
```diff
@@ -2391,6 +2391,194 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    w_ = weights.squeeze()[:, np.newaxis, np.newaxis]
+    # normalize by n_epochs/sum(weights), so that averaging
+    # will yield a correct result
+    epochs._data *= n_epochs * w_ / np.sum(w_)
+
+
+def chpi_weighted_average(raw_snr, raw_epochs=None, reject=False, flat=False,
+                          nharm=None, epoch_start=None, epoch_end=None,
+                          stim_channel=None, mask=None):
+    """ Average epochs weighted by SNR of continuous HPI (cHPI).
```
kambysese: I am not quite sure how I feel about two args `raw_snr` and `raw_epochs` essentially referring to the same thing i.e., the raw input file. At least with how I've been using this script as per conversation with @Eric89GXL & @staulu, the raw input file is preprocessed with SSS & movement compensation using `maxwell_filter`. In this case `raw_snr` and `raw_epochs` seem redundant, and `raw_epochs` could foreseeably refer to epoch `.fif` files with cHPI data intact .

bin/chpi_weighted_average.py
```diff
@@ -0,0 +1,110 @@
...
+                        help=help_nharm)
+    parser.add_argument('--epoch_start', type=float, default=None,
+                        help=help_epoch_start)
+    parser.add_argument('--epoch_end', type=float, default=None,
+                        help=help_epoch_end)
+    parser.add_argument('--plot_snr', help=help_plot_snr, action='store_true')
+    parser.add_argument('--stim_channel', help=help_stim_channel, type=str,
+                        default=None)
+    parser.add_argument('--mask', type=int, default=None, help=help_mask)
+
```
kambysese: @jjnurminen I am liking the additions made here, but in general I think we need more flexibility to allow for integration into the MNEFUN pipeline by hooking into `params` class which defines e.g., SSS, artifact rejection, and averaging conditions etc for the input file. @Eric89GXL WDYT?

mnefun/_mnefun.py
```diff
@@ -2391,6 +2391,194 @@ def plot_reconstruction(evoked, origin=(0., 0., 0.04)):
...
+    epoch_start : float
+        Start of epoch relative to the zero time. If not specified, taken from
+        the data acquisition settings.
+    epoch_end : float
+        End of epoch relative to the zero time. If not specified, taken from
+        the data acquisition settings.
+    stim_channel : str
+        The trigger channel to get the events from.
+    mask : int | None
+        The digital mask to apply to the trigger channel values.
```
Eric89GXL: This is not DRY, and also inconsistent with MNE-Python use cases. This break will make many things difficult, including inclusing in the greater `mnefun` processing pattern. A better plan would be to have the user pass their own `events` array

mnefun/_mnefun.py
Eric89GXL: right, but MaxFilter (tm) does by default (unless you tell it not to), so I think the docstring is okay

mnefun/_mnefun.py
Eric89GXL: Now that I think about it, actually all that needs to be passed in is an `epochs` instance, plus the `raw_chpi` instance. Then all this redundant/non-DRY stuff relating to events can be thrown out. The `epochs` instance has the event timing information within it that can be used to index into `raw_chpi`.

mnefun/_mnefun.py
Eric89GXL: Not in this function. This function should be general. Wherever we call it in `_mnefun.py` is where we should pass in `param` structure stuff.

jjnurminen:
I'm not really familiar with the i-LABS specific mnefun infrastructure (`params` etc.) Of course I'm writing this stuff for you, so I'm perfectly willing to support that also. I'm just thinking that it would be great if the core functionality of this could be kept as generic as possible, so that our lab (and maybe others) could also use it. Do you think that the interfacing to i-LABS pipeline specific things could be accomplished by e.g. some simple wrapper methods?

mnefun/_mnefun.py
jjnurminen: Currently the code just picks `reject` and `flat` out of the acquisition settings, but it's trivial to support them as kwargs also.

mnefun/_mnefun.py
jjnurminen: I've been using this with the typical MaxFilter workflow, i.e. I have the original raw file with cHPI signals intact, and a MaxFiltered version of the same file (tSSS and movement compensation done). Then the idea is to pick the cHPI signals from the original file, and the filtered epochs from the MaxFiltered file, so you need two separate raw file arguments.

If I get @Eric89GXL correctly, you would like to feed already epoched (filtered) data to the method, and it would extract the cHPI epochs from raw data accordingly. I guess the need for that is somehow related to your specific epoching procedures, because otherwise it's simpler just to feed the two Raw instances to it, right?

Eric89GXL:
> If I get @Eric89GXL correctly, you would like to feed already epoched (filtered) data to the method, and it would extract the cHPI epochs from raw data accordingly.

Yeah that's the idea.

> I guess the need for that is somehow related to your specific epoching procedures, because otherwise it's simpler just to feed the two Raw instances to it, right?

Thinking about the cHPI-based averaging algorithm itself, it is actually quite general. It only requires individual epochs, plus something to estimate the cHPI SNR (here a raw instance) of those epochs. So from an implementation standpoint it is already cleaner (orthogonality).

What you have here is potentially simpler for some use cases where users are starting from scratch with raw data. However:

1. Most MNE-Python users (us included) already get their data to the `Epochs` stage using some set of procedures often specific to the paradigm in question. In this sense, doing epoching within the class doesn't help them.
2. From a one-right-way-to-do-things/DRY perspective, it's detrimental to have the same functionality (getting from `Raw` to `Epochs`) implemented multiple places if it doesn't need to be.
3. On a related note, the simplified epoching functionality here only provides a fraction of what MNE-Python allows in terms of event coding/re-encoding, trial-count equalization, decimation, baselining options, etc. For example, even the forced use of `AcqParserFIF` disqualifies it from use on other systems, like BabyMEG, Artemis, or even CTF, all of which have cHPI capabilities. And in our case, extensive event re-encoding (numbers, meaning, even timing) can occur between the original triggers and the ones used for averaging by means of what we call event "scoring".

For something like the shell script, I agree that a workflow that passes in `raw` and a category could make sense, because command-line utilities by design must be simpler than their functional counterparts

jjnurminen:
>On a related note, the simplified epoching functionality here only provides a fraction of what MNE->Python allows in terms of event coding/re-encoding, trial-count equalization, decimation, baselining >options, etc.

Ok, makes sense. However if the epochs have been obtained in some very convoluted way (event recoding and whatever), it looks like the class would have to play the same tricks in order to correctly epoch the cHPI raw data. Or maybe the `Epochs` instance will have all the necessary info? Anyway, will continue from here next week, it's getting late in Finland...

Eric89GXL:
epochs.events should contain the necessary timing information

kambysese:
@jjnurminen Is there a WIP branch on your end for this PR? I'd like to give it a go and can start up a branch if you have more pressing commitments LMK.

jjnurminen:
@kambysese this is the wip branch. I haven't got any unpushed work on it yet.

If you want to implement the approach proposed by @Eric89GXL (feed epochs to the function instead of raw) you can pick it up. Ping me when you're done and I can participate in the review.

jjnurminen:
fixed the conflict with the master branch

larsoner:
Closing this for now due to inactivity, can resume later if need be

PR #155, opened 2016-11-07
WIP: add function for head movement statistics and visualization

ktavabi:
@Eric89GXL I'd like to add a couple of switches `params.plot_head_movement=False` to write out following figures:

_Histogram for head displacement in time as euclidean distance between successive translation vectors from pos file_
hist

_Head movement and rotation produced by porting matlab code @staulu to python_
mvmnt

Figures can later be added to `mne.report`. And also in a separate PR do the same for plotting cHPI SNR. LMWYT

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -76,6 +76,7 @@ from mne.viz import plot_drop_log, tight_layout
 from mne.utils import run_subprocess
 from mne.report import Report
 from mne.io.constants import FIFF
+from mne.transforms import _angle_between_quats
 
 from ._paths import (get_raw_fnames, get_event_fnames, get_report_fnames,
                      get_epochs_evokeds_fnames, safe_inserter, _regex_convert)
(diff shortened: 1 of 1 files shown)


Eric89GXL:
Looks great. Suggestions:

The positions are irregularly sampled in time so velocities are perhaps
better than distances.

PDF of distance from device origin (norm of location basically) would
however be useful. np.bincount with weight equal to duration will get you
this.

Also MNE has an angle between quats function we can use to get angular
velocities.

We can figure out what reasonable plausible maxima are for fast movements
and put those in the plots with e.g. dashed red vertical lines. This gives
point of reference.

This should also end up in MNE, but we can iterate here first. Core
function should only need to take the position quat array.

mnefun/_mnefun.py
```diff
@@ -2561,3 +2563,92 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics=None):
...
+    Parameters
+    ----------
+    rot : array
+
+    """
+    beta = -np.arcsin(rot[2, 0])
+    cosb = np.cos(beta)
+    alpha = np.arcsin(rot[2,1]/cosb)
+    gamma = np.arccos(rot[0,0]/cosb)
+    return alpha, beta, gamma
```
Eric89GXL: use `mne.transforms._angle_between_quats` to get delta angle

mnefun/_mnefun.py
```diff
@@ -2561,3 +2563,92 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics=None):
...
+    title_fontsize = 10
+    tick_fontsize = 10
+    label_fontsize = 10
+
+    # histogram of head displacement data
+    euc_dist *= 100
+    mu = euc_dist.mean()
+    sigma = euc_dist.std()
+    fig = plt.figure()
+    n, bins, patches = plt.hist(euc_dist, 100, normed=1,
```
Eric89GXL: use `np.histogram` with `weights` argument equal to duration of each position

mnefun/_mnefun.py
```diff
@@ -2561,3 +2563,92 @@ def plot_chpi_snr_raw(raw, win_length, n_harmonics=None):
...
+        r'$\mathrm{Head\ displacement:}\ \mu=%.2f cm,\ \sigma=%.2f$' % (
+        mu, sigma))
+    #plt.axis([min(bins), max(bins), min(n), max(n)])
+    plt.grid(True)
+    plt.show()
+
+    # plot head movement
+    fig, ((ax1, ax2), (ax3, ax4), (ax5, ax6)) = plt.subplots(3, 2, sharex=True,
+                                                             figsize=(20, 10))
+    ax1.plot(times, translations[:, 0] * 100)
```
Eric89GXL: use a rcParams context manager to set the title, tick, etc. font sizes

http://matplotlib.org/api/matplotlib_configuration_api.html#matplotlib.rc_context

kambysese:
> PDF of distance from device origin (norm of location basically) would
however be useful. np.bincount with weight equal to duration will get you
this.

WDYM by this? For each location `np.linalg.norm(xyz)` gives distance from device origin. Assuming that is correct, np.bincount with weight equal to duration means what exactly? E.g., 

```
weights = np.unique(np.diff(times)[0]) * np.ones(times.shape)
nbins = np.bincount(poss_norm.astype(int), weights=weights)
```
gives ([   1.,    0.,    0.,    0.,    0.,  341.]) frequencies for normed head positions array, which looks like...

screen shot 2017-03-02 at 10 41 46 pm

Eric89GXL:
Sorry I meant `np.histogram`. Something like:
```
weights = np.diff(times)
pos_norm = np.linalg.norm((pos[1:] + pos[:-1]) / 2., axis=-1)
hist, edges = np.histogram(pos_norm, weights=weights)
```
And you'll probably want to set the number of bins to something reasonable (not sure what that would be, needs experimenting).

Eric89GXL:
BTW I have a PR open to MNE-Python to the position as a function of time plotting. I think we could just add this histogram mode to that function, as well.

larsoner:
@kambysese okay with you if I take this one over?

kambysese:
Yeah :disappointed: Sorry

larsoner:
@ktavabi I'm assuming we don't need this now that we have head position plotting, but if you still want it I can take over

ktavabi:
I thought you had taken this over as of Oct! No I don't think we need this, in particular, rather a PR to add information on rotation and mov velocities.

PR #156, opened 2016-11-09
FIX:minor invalid arg to epochs.equalize_event_counts

kambysese:
@Eric89GXL this looks to be a conflict


Eric89GXL:
Yeah that argument is gone in the latest `master`

PR #163, opened 2016-11-18
Fix plot psd

### LINKED ISSUE #158: List input to Raw causes AttributeError
`plot_raw_psd` line 2237 causes AttributeError b/c `fname` arg to `Raw` is a list instead of filename `str`. @Eric89GXL I will fix in PR.

kambysese:
Fixes #158


Eric89GXL:
Thanks @kambysese

PR #164, opened 2016-12-14
fix indexing bug

jjnurminen:
This fixes a "double indexing" bug, where MEG channels were already picked into a 306-dim matrix and then again indexed with `pick_grad` and `pick_mag`. It manifested as an IndexError in a case where the fiff channel list did not start with MEG channels.


jjnurminen:
(also changed some `.transpose()` calls to `.T`, not related to the bug)

Eric89GXL:
LGTM, ready to go from your end?

jjnurminen:
I'm ready

Eric89GXL:
Thanks for the fix

PR #167, opened 2017-03-07
Make plots

kambysese:
@Eric89GXL maybe go with this for now for generating head movement plot. WDYT?

### DIFF
--- a/examples/funloc/analysis_fun.py
+++ b/examples/funloc/analysis_fun.py
@@ -77,7 +77,7 @@ params.proj_nums = [[1, 1, 0],  # ECG
                     [1, 1, 2],  # EOG
                     [0, 0, 0]]  # Continuous (from ERM)
 params.cov_method = 'shrunk'  # Cleaner noise covariance regularization
-
+params.plot_head_position = True  # Plot cHPI data for single raw file
 # python | maxfilter for choosing SSS applied using either Maxfilter or MNE
 params.sss_type = 'python'
 # The scoring function needs to produce an event file with these values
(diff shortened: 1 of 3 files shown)


Eric89GXL:
That looks like a good approach, except that it only plots the first run's head movement. I think it's fine to plot them separately or together, up to you.

kambysese:
Oh hell! @Eric89GXL I probably should not have rebased before the recent commit? I only want to merge `d5bfc03`. How do I clean this up?

examples/funloc/analysis_fun.py
```diff
@@ -123,6 +123,7 @@
     gen_ssp=True,       # Generate SSP vectors
     apply_ssp=True,     # Apply SSP vectors and filtering
     plot_psd=True,      # Plot raw data power spectra
+    plot_hp=True,       # plot head movement from raw data
```
Eric89GXL: we shouldn't need both a `params.` and a boolean here. Probably just the boolean here would be good

kambysese: Sorry slipped back in after I cleaned up a botched commit history.

mnefun/_mnefun.py
```diff
@@ -239,6 +240,8 @@ class Params(Frozen):
     sss_origin : array-like, shape (3,) | str
         Origin of internal and external multipolar moment space in meters.
         Default is center of sphere fit to digitized head points.
+    plot_hp : bool
+        Plot head movement traces from raw data
```
Eric89GXL: this docstring is out of date. It's not part of the constructor anyway so let's get rid of it

kambysese:
I compared with current master and deleted outdated commits. Should be ready for merge.

Eric89GXL:
Thanks @kambysese

PR #168, opened 2017-04-05
MRG: Updates for latest MNE

Eric89GXL:
Some minor fixes for latest MNE.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -21,6 +21,7 @@ from scipy import io as spio
 import matplotlib.pyplot as plt
 from numpy.testing import assert_allclose
 
+import mne
 from mne import (compute_proj_raw, make_fixed_length_events, Epochs,
                  find_events, read_events, write_events, concatenate_events,
                  read_cov, compute_covariance, write_cov,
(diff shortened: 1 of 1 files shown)


Eric89GXL:
@kambysese do these work for you, too?

kambysese:
With the exception of `fir_design` kwarg, yes the updates have worked for me through `write_epochs`

mnefun/_mnefun.py
```diff
@@ -414,9 +416,22 @@ def convert_subjects(self, subj_template, struc_template=None):
...
+            assert all(isinstance(subj, string_types) for subj in new)
+            self.structurals = new
+        if isinstance(subj_template, string_types):
+            def fun(x):
+                return subj_template % x
+        else:
+            fun = subj_template
+        new = [fun(subj) for subj in self.subjects]
+        assert all(isinstance(subj, string_types) for subj in new)
+        self.subjects = new
```
kambysese: Is this bit for inverse modelling in a template_mri?

Eric89GXL: It allows more flexible mapping between `subj` and `struct` names, e.g. if you need to put in a `subj.upper()` for example

Eric89GXL:
Okay @kambysese this should work with old and new MNE now

mnefun/_mnefun.py
```diff
@@ -1706,6 +1707,21 @@ def _cals(raw):
...
+def _get_fir_kwargs(fir_design):
+    """Get FIR kwargs in backward-compatible way."""
+    fir_kwargs = dict()
+    if fir_design != 'firwin2':
+        # only pass if necessary (backward compat)
+        fir_kwargs.update(fir_design=fir_design)
+    old_kwargs = dict()
+    if 'fir_design' in mne.fixes._get_args(mne.filter.filter_data):
+        old_kwargs.update(fir_design='firwin2')
+    elif len(fir_kwargs):
```
kambysese: What's happening here? Isn't `elif len(fir_kwargs)` always going evaluate as True?

Eric89GXL: No, only if `fir_design != 'firwin2'`

kambysese:
Apart from my question LTGM

mnefun/_mnefun.py
kambysese: Even though it's not indented into the `fir_design != 'firwin2' ` block? sorry I don't get it...

Eric89GXL: You'll have to think about the various cases, i.e. the values of `fir_design` and the MNE version, and think about the flow. But in any case, I think it should work okay

Eric89GXL:
Thanks @kambysese. Let me know if I broke anything when you actually use it :)

PR #171, opened 2017-04-14
FIX default drop_logs plots behavior

kambysese:
Prevent drop_logs plots windows popping up after  do_preprocessing in large batch jobs


Eric89GXL:
Thanks @kambysese

PR #174, opened 2017-05-31
FIX mne setup_source_spaces version conflict

kambysese:
@Eric89GXL filename arg is deprecated and causing issues in `mne.setup_source_space`

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1594,7 +1594,7 @@ def gen_forwards(p, subjects, structurals, run_indices):
                                      structurals[si] + '-oct-6-src.fif')
             if not op.isfile(src_space_file):
                 print('  Creating source space for %s...' % subj)
-                src = setup_source_space(structurals[si], 'oct6',
+                src = setup_source_space(structurals[si], spacing='oct6',
                                          n_jobs=p.n_jobs)
                 write_source_spaces(src_space_file, src)
             src = read_source_spaces(src_space_file)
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -1589,12 +1590,14 @@ def gen_forwards(p, subjects, structurals, run_indices):
...
-            src = op.join(subjects_dir, structurals[si], 'bem',
-                          structurals[si] + '-oct-6-src.fif')
-            if not op.isfile(src):
+            src_space_file = op.join(subjects_dir, structurals[si], 'bem',
+                                     structurals[si] + '-oct-6-src.fif')
+            if not op.isfile(src_space_file):
                 print('  Creating source space for %s...' % subj)
-                setup_source_space(structurals[si], src, 'oct6',
-                                   n_jobs=p.n_jobs)
+                src = setup_source_space(structurals[si], 'oct6',
```
Eric89GXL: better to pass `spacing='oct6'` to help the transition

kambysese: Will do. I was actually thinking of making spacing a Params arg, WDYT?

Eric89GXL: Do you need to change it from the recommended default?

kambysese: Not me. But I also never thought I'd be doing MEG in populations from the recommended default. I'll leave it as is.

Eric89GXL: YAGNI

kambysese: Done

Eric89GXL:
Thanks @kambysese

PR #175, opened 2017-05-31
ENH allow user defined Neuromag site files

### LINKED ISSUE #173: Neuromag cal files
@Eric89GXL you ok with allowing `params.cal_file` & `params.ct_file` args for user defined neuromag site files?

kambysese:
Fixes #173

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1004,11 +1004,13 @@ def run_sss_locally(p, subjects, run_indices):
     mne.preprocessing.maxwell_filter
     """
     data_dir = op.join(op.dirname(__file__), 'data')
-    if p.cal_file and p.ct_file is None:
+    if p.cal_file is None:
         cal_file = op.join(data_dir, 'sss_cal.dat')
-        ct_file = op.join(data_dir, 'ct_sparse.fif')
     else:
         cal_file = p.cal_file
+    if p.ct_file is None:
+        ct_file = op.join(data_dir, 'ct_sparse.fif')
+    else:
         ct_file = p.ct_file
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -1002,8 +1004,12 @@ def run_sss_locally(p, subjects, run_indices):
     mne.preprocessing.maxwell_filter
     """
     data_dir = op.join(op.dirname(__file__), 'data')
-    cal_file = op.join(data_dir, 'sss_cal.dat')
-    ct_file = op.join(data_dir, 'ct_sparse.fif')
+    if p.cal_file and p.ct_file is None:
```
Eric89GXL: I think you mean `if p.cal_file is None and p.ct_file is None` but really they should be separate conditionals

kambysese: Yes I meant `if p.cal_file is None and p.ct_file is None` somehow what I have is working :confused:  I thought to combine the conditional since both files need to be provided. 

You mean like....
```
    if p.cal_file is None:
        cal_file = op.join(data_dir, 'sss_cal.dat')
    else:
        cal_file = p.cal_file
    if p.ct_file is None:
        ct_file = op.join(data_dir, 'ct_sparse.fif')
    else:
        ct_file = p.ct_file
```
seems a little bulky, no?

Eric89GXL:
> somehow what I have is working

Yes what you have will work in some/most cases but it's not as clean

> seems a little bulky, no?

It covers cases your code doesn't currently. For example, if one passes `ct_fname=None` and `cal_file='...'` your code will not use any cross-talk correction, but if one passes `ct_fname=None` and `cal_file=None` it will.

Eric89GXL:
Thanks @kambysese

PR #180, opened 2017-08-31
FIX: Fix proj

### LINKED ISSUE #179: PCA procedure
@larsoner I noticed that even when all `params.proj_nums` are set to `[0, 0, 0]` it looks like `do_preprocessing_combined` still computes PCA. Is that normal/expected behavior?

larsoner:
Closes #179.

@kambysese can you see if this fixes your problem?

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -658,8 +658,8 @@ def fetch_raw_files(p, subjects, run_indices):
         want = set(op.basename(fname) for fname in fnames)
         got = set(op.basename(fname) for fname in remote_fnames)
         if want != got.intersection(want):
-            raise RuntimeError('Could not find all files.\nWanted: %s\nGot: %s'
-                               % (want, got.intersection(want)))
+            raise RuntimeError('Could not find all files, missing:\n' +
+                               '\n'.join(sorted(want - got)))
         if len(remote_fnames) != len(fnames):
             warnings.warn('Found more files than expected on remote server.\n'
                           'Likely split files were found. Please confirm '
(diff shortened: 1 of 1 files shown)


kambysese:
Nope. Threw...

>   File "/home/ktavabi/Projects/ibis/ibis/preprocess.py", line 115, in <module>
>     print_status=True
>   File "/home/ktavabi/Projects/mnefun/mnefun/_mnefun.py", line 614, in do_processing
>     outs[ii] = func(p, subjects, run_indices)
>   File "/home/ktavabi/Projects/mnefun/mnefun/_mnefun.py", line 1925, in do_preprocessing_combined
>     if projs.shape != (3, 3):
> AttributeError: 'list' object has no attribute 'shape'

`projs` is a list at ln 1925 and even if the number of desired components is [0, 0, 0] for the different projection types projs[0].shape is always 3-by-3. Or am I missing something?

larsoner:
Should be fixed now.

You shouldn't pass `[0, 0, 0]` but `[[0] * 3] * 3]` or any equivalent. If you want a shortcut, we'll need to add it.

kambysese:
Okay, working now. Not sure how bc I've been passing [[0, 0, 0], [0, 0, 0], [0, 0, 0]], not just [0, 0, 0] and that list is shape 3x3.

larsoner:
Your report from #179 was that you used `[0, 0, 0]`, which is the wrong shape

larsoner:
(`[[0, 0, 0], [0, 0, 0], [0, 0, 0]]` should already work in `master`)

kambysese:
I stand corrected. Head taking penalty shots at my arse. Sorry to waste your time.

larsoner:
It's alright, we'll at least get a more reasonable error now

PR #184, opened 2017-09-20
ENH:add autoreject channel thresholds to epoching

ktavabi:
@larsoner please review. And do you think you could test on funloc sample data once we agree on implementation?

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -44,7 +44,6 @@ from mne.preprocessing.maxwell import (maxwell_filter,
                                        _get_mf_picks, _prep_mf_coils,
                                        _check_regularize,
                                        _regularize)
-from mne.preprocessing import create_ecg_epochs, create_eog_epochs
 from mne.utils import verbose
 
 try:
@@ -619,8 +618,6 @@ def do_processing(p, fetch_raw=False, do_score=False, push_raw=False,
                 outs[ii] = func(p, subjects, p.in_names, p.in_numbers,
                                 p.analyses, p.out_names, p.out_numbers,
                                 p.must_match, decim, run_indices)
-            elif func == do_preprocessing_combined:
-                outs[ii] = func(p, subjects, run_indices, decim)
             elif func == print_proc_status:
                 outs[ii] = func(p, subjects, structurals, p.analyses,
                                 run_indices)
@@ -1428,10 +1425,6 @@ def save_epochs(p, subjects, in_names, in_numbers, analyses, out_names,
                                  preload=True, decim=decim[si])
             new_dict = get_rejection_threshold(temp_epochs)
             use_reject, use_flat = _restrict_reject_flat(new_dict, p.flat, raw)
-            new_dict = get_rejection_threshold(temp_epochs)
-            use_reject = {k: new_dict[k]
-                          for k in ['grad', 'mag', 'eeg']
-                          if k in new_dict.keys()}
         else:
             use_reject, use_flat = _restrict_reject_flat(p.reject, p.flat, raw)
 
@@ -1841,7 +1834,7 @@ def _raw_LRFCP(raw_names, sfreq, l_freq, h_freq, n_jobs, n_jobs_resample,
     return raw
 
 
-def do_preprocessing_combined(p, subjects, run_indices, decim):
+def do_preprocessing_combined(p, subjects, run_indices):
     """Do preprocessing on all raw files together
 
     Calculates projection vectors to use to clean data.
@@ -1854,11 +1847,7 @@ def do_preprocessing_combined(p, subjects, run_indices, decim):
         Subject names to analyze (e.g., ['Eric_SoP_001', ...]).
     run_indices : array-like | None
         Run indices to include.
-    decim : int | list of int
-        Amount to decimate.
     """
-    if p.autoreject_thresholds:
-        from autoreject import get_rejection_threshold
     drop_logs = list()
     for si, subj in enumerate(subjects):
         if p.disp_files:
@@ -1906,7 +1895,7 @@ def do_preprocessing_combined(p, subjects, run_indices, decim):
             epochs = Epochs(raw, events, None, p.tmin, p.tmax,
                             baseline=_get_baseline(p), picks=picks,
                             reject=p.auto_bad_reject, flat=p.auto_bad_flat,
-                            proj=True, preload=True, decim=decim[si],
+                            proj=True, preload=True, decim=1,
                             reject_tmin=p.reject_tmin,
                             reject_tmax=p.reject_tmax)
             # channel scores from drop log
@@ -2033,17 +2022,6 @@ def do_preprocessing_combined(p, subjects, run_indices, decim):
                        **old_kwargs)
             raw.add_proj(projs)
             raw.apply_proj()
-            if p.autoreject_thresholds:
-                ecg_epochs = create_ecg_epochs(raw, tmin=ecg_t_lims[0],
-                                               tmax=ecg_t_lims[1],
-                                               ch_name=p.ecg_channel)
-                new_dict = get_rejection_threshold(ecg_epochs)
-                # only reject ecg events for MEG channels for now
-                ecg_reject = {k: new_dict[k]
-                              for k in ['grad', 'mag',
-                                        'eeg'] if k in new_dict.keys()}
-            else:
-                ecg_reject = p.ssp_ecg_reject
             pr, ecg_events = \
                 compute_proj_ecg(raw, n_grad=proj_nums[0][0],
                                  n_jobs=p.n_jobs_mkl,
@@ -2051,7 +2029,7 @@ def do_preprocessing_combined(p, subjects, run_indices, decim):
                                  tmin=ecg_t_lims[0], tmax=ecg_t_lims[1],
                                  l_freq=None, h_freq=None, no_proj=True,
                                  qrs_threshold='auto', ch_name=p.ecg_channel,
-                                 reject=ecg_reject)
+                                 reject=p.ssp_ecg_reject)
             if ecg_events.shape[0] >= 20:
                 write_events(ecg_eve, ecg_events)
                 write_proj(ecg_proj, pr)
@@ -2075,17 +2053,6 @@ def do_preprocessing_combined(p, subjects, run_indices, decim):
                        **old_kwargs)
             raw.add_proj(projs)
             raw.apply_proj()
-            if p.autoreject_thresholds:
-                eog_epochs = create_eog_epochs(raw, tmin=eog_t_lims[0],
-                                               tmax=eog_t_lims[1],
-                                               ch_name=p.ecg_channel)
-                new_dict = get_rejection_threshold(eog_epochs)
-                # only reject ecg events for MEG channels for now
-                ecg_reject = {k: new_dict[k]
-                              for k in ['grad', 'mag',
-                                        'eeg'] if k in new_dict.keys()}
-            else:
-                ecg_reject = p.ssp_ecg_reject
             pr, eog_events = \
                 compute_proj_eog(raw, n_grad=proj_nums[1][0],
                                  n_jobs=p.n_jobs_mkl,
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -78,11 +78,6 @@ from mne.utils import run_subprocess
 from mne.report import Report
 from mne.io.constants import FIFF
 
-try:
-    from autoreject import get_rejection_threshold
-except ImportError, e:
-    print(e)
-
 from ._paths import (get_raw_fnames, get_event_fnames, get_report_fnames,
                      get_epochs_evokeds_fnames, safe_inserter, _regex_convert)
 from ._status import print_proc_status
@@ -251,7 +246,8 @@ class Params(Frozen):
         Can be "firwin2" or "firwin".
     autoreject_thresholds : bool | False
         If True use autoreject module to compute global rejection thresholds
-        for epoching.
+        for epoching. Make sure autoreject module is installed. See
+        http://autoreject.github.io/ for instructions.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


mnefun/_mnefun.py
```diff
@@ -77,6 +78,11 @@
 from mne.report import Report
 from mne.io.constants import FIFF
 
+try:
+    from autoreject import get_rejection_threshold
+except ImportError, e:
+    print(e)
```
larsoner: Nest import to `autoreject` in whatever nested clause it will actually be used. Then no need for `try/except`

mnefun/_mnefun.py
```diff
@@ -1380,7 +1390,21 @@ def save_epochs(p, subjects, in_names, in_numbers, analyses, out_names,
                 warnings.warn('resulting new sampling frequency %s not equal '
                               'to previous values %s' % (new_sfreq, sfreqs))
             sfreqs.add(new_sfreq)
-        use_reject, use_flat = _restrict_reject_flat(p.reject, p.flat, raw)
+        _, use_flat = _restrict_reject_flat(p.reject, p.flat, raw)
```
larsoner: This should go inside the `if p.autoreject_thresholds:`

mnefun/_mnefun.py
```diff
@@ -1380,7 +1390,21 @@ def save_epochs(p, subjects, in_names, in_numbers, analyses, out_names,
...
+            print('     Using autreject to compute rejection thresholds')
+            temp_epochs = Epochs(raw, events, event_id=None, tmin=p.tmin,
+                                 tmax=p.tmax, baseline=_get_baseline(p),
+                                 proj=True, reject=None, flat=use_flat,
+                                 preload=True)
+            new_dict = get_rejection_threshold(temp_epochs)
+            # only reject ecg events for MEG channels for now
+            #TODO test if ExG thresholds work during epoching
+            use_reject = {k: new_dict[k]
+                          for k in ['grad', 'mag'] if k in new_dict.keys()}
```
larsoner: `['grad', 'mag', 'eeg']`

mnefun/_mnefun.py
```diff
@@ -1824,6 +1848,7 @@ def do_preprocessing_combined(p, subjects, run_indices):
                   '        %s' % bad_file)
             if not op.isdir(bad_dir):
                 os.mkdir(bad_dir)
+            #TODO 86 autobad after autoreject merge
```
larsoner: what does this mean?

mnefun/_mnefun.py
```diff
@@ -1888,7 +1913,7 @@ def do_preprocessing_combined(p, subjects, run_indices):
 
         proj_nums = p.proj_nums
         eog_t_lims = [-0.25, 0.25]
-        ecg_t_lims = [-0.08, 0.08]
+        ecg_t_lims = [-0.2, 0.2]
```
larsoner: why this change? It should probably be a parameter if it needs tweaking, as this is a backward-incompatible change

kambysese:
I'll get to the rest of this tomorrow. I must have missed the email notification about your review. Thanks.

mnefun/_mnefun.py
kambysese: Thinking we could get rid of the automatic bad channel detection relying on arbitrary thresholds since autoreject module appears to also be able to "fix" bad channels.

larsoner: We should leave all old methods present/functional for backward compatibility

mnefun/_mnefun.py
kambysese: I was just playing around with this and forgot to change it back. I was curious as to why you went with such a narrow window for QRS epoching? In the mne tutorials, the window is -500 - 500 ms and I've been using -200 - 200 ms for infants. Why is this backward-incompatible?

larsoner: >  was curious as to why you went with such a narrow window for QRS epoching?

I don't remember but I doubt it makes a huge difference

> Why is this backward-incompatible?

It will change the computations and thus the results you get if you run `mnefun` before / after this PR (and not in a bug-fixing sort of way)

kambysese:
Revised as per comments. Tested on funloc subject data...

image

yellow = epoching with `proj=True` + autoreject
red = epoching with `proj=False`  + autoreject
cyan = epoching with `proj=False` without autoreject

I am thinking we should set `proj=True` WDYT. I just noticed there is a conflict. Do you want to review and then I'll try to resolve before merge?

larsoner:
> I am thinking we should set proj=True WDYT.

I think we should always use `proj='delayed'`, that way we can disable projs later if we want to.

mnefun/_mnefun.py
```diff
@@ -1380,7 +1386,22 @@ def save_epochs(p, subjects, in_names, in_numbers, analyses, out_names,
...
             sfreqs.add(new_sfreq)
-        use_reject, use_flat = _restrict_reject_flat(p.reject, p.flat, raw)
+        if p.autoreject_thresholds:
+            from autoreject import get_rejection_threshold
+            print('     Using autreject to compute rejection thresholds')
+            temp_epochs = Epochs(raw, events, event_id=None, tmin=p.tmin,
+                                 tmax=p.tmax, baseline=_get_baseline(p),
+                                 proj=False, reject=None, flat=None,
+                                 preload=True)
+            new_dict = get_rejection_threshold(temp_epochs, decim=2)
```
larsoner: don't use `decim=2` here, use `decim[si]` in the `Epochs` constructor

mnefun/_mnefun.py
```diff
@@ -1969,14 +1990,24 @@ def do_preprocessing_combined(p, subjects, run_indices):
                        **old_kwargs)
             raw.add_proj(projs)
             raw.apply_proj()
+            if p.autoreject_thresholds:
+                ecg_epochs = create_ecg_epochs(raw, tmin=ecg_t_lims[0],
+                                               tmax=ecg_t_lims[1],
+                                               ch_name=p.ecg_channel)
+                new_dict = get_rejection_threshold(ecg_epochs, decim=3)
```
larsoner: same comment here

mnefun/_mnefun.py
kambysese:
```
...
    elif func == do_preprocessing_combined:
        if p.autoreject_thresholds:
            from autoreject import get_rejection_threshold
        outs[ii] = func(p, subjects, run_indices)
....
```

For the sake of not DRY'ing why does above snippet not work in `do_processing`. I can see the code go into the conditional and executing ` if p.autoreject_thresholds` as `True` but I still get 

> NameError: global name 'get_rejection_threshold' is not defined

down in `do_processing_combined`

larsoner: It's a scope problem. When you import `get_rejection_threshold` in that function, the `func` (whatever it is, which must be `do_processing_combined`) does not have access to it. You should do the import in `do_processing_combined`.

larsoner: (A good introspecting editor will give you a flake error in both functions, saying that `get_rejection_threshold` isn't used in the first function, and isn't known in the second)

kambysese: Thanks. Pycharm has no native support for flake :disappointed:

kambysese:
I am smelling a successful conflict resolution after two years of git --inon
Good to merge?

mnefun/_mnefun.py
```diff
@@ -1406,7 +1414,23 @@ def save_epochs(p, subjects, in_names, in_numbers, analyses, out_names,
                 warnings.warn('resulting new sampling frequency %s not equal '
                               'to previous values %s' % (new_sfreq, sfreqs))
             sfreqs.add(new_sfreq)
-        use_reject, use_flat = _restrict_reject_flat(p.reject, p.flat, raw)
+        if p.autoreject_thresholds:
+            if p.autoreject_thresholds:
```
larsoner: you do not need this second conditional

mnefun/_mnefun.py
```diff
@@ -1406,7 +1414,23 @@ def save_epochs(p, subjects, in_names, in_numbers, analyses, out_names,
...
+                from autoreject import get_rejection_threshold
+            print('     Using autreject to compute rejection thresholds')
+            temp_epochs = Epochs(raw, events, event_id=None, tmin=p.tmin,
+                                 tmax=p.tmax, baseline=_get_baseline(p),
+                                 proj=True, reject=None, flat=None,
+                                 preload=True)
+            new_dict = get_rejection_threshold(temp_epochs, decim=decim[si])
```
larsoner: do not pass `decim=decim[si]` here but rather in the `Epochs` constructor

kambysese:
There is an algorithm to [find and repair bad channels](http://autoreject.github.io/generated/autoreject.Ransac.html#autoreject.Ransac) but I've not really played around with it. I am okay to merge global thresholding for now. It would be nice if there were some feedback from others on this stuff e.g., how things are working out for them.

kambysese:
@larsoner Can we finish this too while we at it?

larsoner:
Merge conflict

kambysese:
Happy?

mnefun/_mnefun.py
```diff
@@ -1411,7 +1419,22 @@ def save_epochs(p, subjects, in_names, in_numbers, analyses, out_names,
...
+            print('     Using autreject to compute rejection thresholds')
+            temp_epochs = Epochs(raw, events, event_id=None, tmin=p.tmin,
+                                 tmax=p.tmax, baseline=_get_baseline(p),
+                                 proj=True, reject=None, flat=None,
+                                 preload=True, decim=decim[si])
+            new_dict = get_rejection_threshold(temp_epochs)
+            use_reject, use_flat = _restrict_reject_flat(new_dict, p.flat, raw)
+            new_dict = get_rejection_threshold(temp_epochs)
+            use_reject = {k: new_dict[k]
+                          for k in ['grad', 'mag', 'eeg']
```
larsoner: Should this also have `eog` and `ecg` in it? I think our `reject` can usually have at least `eog`, but I'm not sure if `autoreject` does anything with the non-M/EEG channels

kambysese: So what happens if I use the entire rejection dictionary from autoreject e.g. `{'ecg': 2.93e-12, 'eog': 9.1e-05, 'eeg': 9.1e-05, 'grad': 7.97e-11, 'mag': 2.93e-12}` all epochs are rejected b/c ExG channels always exceed the autoreject criteria. In the case funloc subj_01 ECG063 causes all trials to be dropped. I assume this is because of how autoreject deals with ExG channels.

larsoner: This sounds like a bug with `autoreject`. Can you replicate the problem with the MNE sample dataset?

kambysese: I can run the question by Mainak, and maybe he'll look into it for us. But the other reason I opted for `['grad', 'mag', 'eeg']` is that IIRC in the past you have argued for using PCA to project out ExG artifacts rather than rejecting epochs based on their amplitudes.

larsoner: We'll have to talk to the `autoreject` folks about how these thresholds are chosen. We do want to keep the artifacts of interest, but not e.g. gigantic channel discontinuity artifacts. It's possible that even using `eeg` here would try to omit most of the blink epochs, for example, depending on how `autoreject` works.

kambysese:
WDYM by
"We do want to keep the artifacts of interest"
I posted a question to Mainak on gitter and CCed you.

larsoner:
Blinks and heartbeat events

kambysese:
Right. But what do you mean by keeping them?
I didn't get much of an explanation on gitter, but recent PR excluded EEG from threshold dictionary. What do you make of it?

larsoner:
> But what do you mean by keeping them?

In at least some of the code of interest, we are creating epochs around artifacts (blink, heartbeat) so we want to keep them (not have them be rejected)

> I didn't get much of an explanation on gitter, but recent PR excluded EEG from threshold dictionary. What do you make of it?

What recent PR?

kambysese:
> In at least some of the code of interest, we are creating epochs around artifacts (blink, heartbeat) so we want to keep them (not have them be rejected)

Ok. But that's not happening with `['grad', 'mag', 'eeg']` sub-selection from autoreject dictionary...I don't think.

[This PR](https://github.com/autoreject/autoreject/pull/81#issuecomment-334484599)...sorry markdown is a SOB on the phone.

mnefun/_mnefun.py
```diff
@@ -2037,6 +2075,17 @@ def do_preprocessing_combined(p, subjects, run_indices):
                        **old_kwargs)
             raw.add_proj(projs)
             raw.apply_proj()
+            if p.autoreject_thresholds:
+                eog_epochs = create_eog_epochs(raw, tmin=eog_t_lims[0],
+                                               tmax=eog_t_lims[1],
+                                               ch_name=p.ecg_channel)
+                new_dict = get_rejection_threshold(eog_epochs)
```
larsoner: Here and above we use `autoreject` to set the rejection params. We need to verify that it's not eliminating too many EOG epochs. Have you checked to see what proportion are removed? It should be fairly small for most datasets.

kambysese:
@larsoner Can you review the following gist. In it, I am trying to see what proportions of EOG events are dropped with default rejection values from mne e.g., `dict(grad=4000e-13, mag=4e-12, eeg=500e-6, eog=250e-6)` vs. `{'eeg': 0.000174, 'eog': 0.000230, 'grad': 9.525e-11, 'mag': 3.01e-12}` from autoreject.

I am surprised that the number of epochs dropped by `compute_proj_eog` and `create_eog_epochs` using the same rejection criteria is not the same. Is that the expected behavior?
Secondly,  it looks like the criteria from autoreject will result in more epochs dropped when considering all channel types. Again there is a discrepancy between the number dropped epochs between `compute_proj_eog` and `create_eog_epochs`, with the latter being 2x as many with autoreject compared to arbitrary criteria. 

```
from mne.datasets.sample import data_path
from mne.io import read_raw_fif
from mne import read_events
from mne.preprocessing import compute_proj_eog, create_eog_epochs
from autoreject import get_rejection_threshold

data_path = data_path()
raw = read_raw_fif(data_path + '/MEG/sample/sample_audvis_filt-0-40_raw.fif',
                   preload=True)
events = read_events(data_path + '/MEG/sample/sample_audvis_filt-0-40_raw-eve.fif')

pr, eog_events_a = compute_proj_eog(raw, n_grad=1,
                                    n_mag=1, n_eeg=1,
                                    tmin=-.25, tmax=.25,
                                    reject=dict(grad=4000e-13,
                                                mag=4e-12,
                                                eeg=500e-6,
                                                eog=250e-6))  # 12 eps drop

eog_epochs = create_eog_epochs(raw, reject=dict(grad=4000e-13,
                                                   mag=4e-12,
                                                   eeg=500e-6,
                                                   eog=250e-6))  # 16 eps drop  # noqa
eog_epochs.average().plot()

print('number of EOG events: %s' % len(eog_events_a))  # 46 events

# with autoreject
reject = get_rejection_threshold(eog_epochs)
print(reject)
pr, eog_events_b = compute_proj_eog(raw, n_grad=1,
                                    n_mag=1, n_eeg=1,
                                    tmin=-.25, tmax=.25,
                                    reject=reject)  # 15 eps drop

eog_epochs = create_eog_epochs(raw, reject=reject)  # 31 eps drop
eog_epochs.average().plot()
print('number of EOG events with autoreject: %s' % len(eog_events_b))  # 46 events
```

larsoner:
> I am surprised that the number of epochs dropped by compute_proj_eog and create_eog_epochs using the same rejection criteria is not the same. Is that the expected behavior?

Looking at the API doc, [compute_proj_eog](http://mne-tools.github.io/dev/generated/mne.preprocessing.compute_proj_eog.html) filtetrs the data before processing and [create_eog_epochs](https://mne-tools.github.io/dev/generated/mne.preprocessing.create_eog_epochs.html) does not. Presumably if you filter `raw` before sending to the latter they will match.

ktavabi:
> Looking at the API doc, compute_proj_eog filtetrs the data before processing and create_eog_epochs does not.

The default filtering args for `compute_proj_eog` are ` l_freq=1.0`, `h_freq=35.0`, and 1.0 & 10, respectively for `create_eog_epochs`. So I am not sure what you mean.

In hindsight tho it make very little sense to use autoreject to compute rejection criteria for artifact epochs, especially considering we want to maximize the number of epochs for input into SSP. I am reverting back to previous state. Here are results from funloc subj02 with red trace using autoreject dict and yellow using mnefun defaults.
autorej_testing

larsoner:
Thanks @ktavabi

PR #185, opened 2017-09-25
WIP:plot artifact SSP topographies and save.

kambysese:
@larsoner currently this WIP plots SSP projection topographies of artifacts for MEG sensors and saves figure in `sss_pca_fif` directory. I have two questions:

1. Does `mne.viz.plot_projs_topomap` already handle multiple projection components across channel types e.g., if I request 2 SSP projectors for Grad and 1 projector for Mags will the program plot accordingly

2. How would you advise dealing with EEG projections? As is `mne.viz.plot_projs_topomap` returns a RT error because it cannot find appropriate EEG layout in funloc data. 

LMKWYT.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1925,6 +1925,9 @@ def do_preprocessing_combined(p, subjects, run_indices):
                 print('    Adding extra projectors from "%s".' % p.proj_extra)
             extra_proj = op.join(pca_dir, p.proj_extra)
             projs = read_proj(extra_proj)
+            if p.plot_pca:
+                h = plot_projs_topomap(projs, info=raw_orig.info, show=False)
+                h.savefig(extra_proj[:-4] + '.png', format='png', dpi=120)
 
         # Calculate and apply ERM projectors
         proj_nums = np.array(proj_nums, int)
@@ -1958,6 +1961,9 @@ def do_preprocessing_combined(p, subjects, run_indices):
                                   reject=None, flat=None, n_jobs=p.n_jobs_mkl)
             write_proj(cont_proj, pr)
             projs.extend(pr)
+            if p.plot_pca:
+                h = plot_projs_topomap(pr, info=raw.info, show=False)
+                h.savefig(cont_proj[:-4] + '.png', format='png', dpi=120)
             del raw
 
         # Calculate and apply the ECG projectors
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -404,7 +404,7 @@ class Params(Frozen):
         self.must_match = []
         self.on_missing = 'error'  # for epochs
         self.subject_run_indices = None
-        self.plot_pca = False
+        self.plot_pca = True
         self.freeze()
 
     @property
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -71,7 +71,7 @@ from mne.io.pick import pick_types_forward, pick_types
 from mne.io.meas_info import _empty_info
 from mne.cov import regularize
 from mne.minimum_norm import write_inverse_operator
-from mne.viz import plot_drop_log, tight_layout
+from mne.viz import plot_drop_log, tight_layout, plot_projs_topomap
 from mne.viz._3d import plot_head_positions
 from mne.utils import run_subprocess
 from mne.report import Report
@@ -243,6 +243,9 @@ class Params(Frozen):
         Default is center of sphere fit to digitized head points.
     fir_design : str
         Can be "firwin2" or "firwin".
+    plot_pca : bool
+        If set to True generate selected PCA component topographies and save
+        figures to disk in pca_fif folder.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


larsoner:
> Does mne.viz.plot_projs_topomap already handle multiple projection components across channel types e.g., if I request 2 SSP projectors for Grad and 1 projector for Mags will the program plot accordingly

Not sure, you'll have to try it out

> How would you advise dealing with EEG projections? As is mne.viz.plot_projs_topomap returns a RT error because it cannot find appropriate EEG layout in funloc data.

It shouldn't, this is a bug. Can you reproduce it with the `sample` dataset and open an MNE issue?

kambysese:
```
data_path = sample.data_path()
raw_fname = data_path + '/MEG/sample/sample_audvis_filt-0-40_raw.fif'
raw = mne.io.read_raw_fif(raw_fname, preload=True)
raw.set_eeg_reference()
raw.pick_types(meg=True, ecg=True, eog=True, stim=True)
projs, events = compute_proj_eog(raw, n_grad=1, n_mag=1, average=True)
print(projs)
# eog_projs = projs[-2:]
mne.viz.plot_projs_topomap(projs)
```

Gives...

>   File "/home/ktavabi/miniconda3/envs/py2.7/lib/python2.7/site-packages/IPython/core/interactiveshell.py", line 2881, in run_code
>     exec(code_obj, self.user_global_ns, self.user_ns)
>   File "<ipython-input-3-dbdf234bbb03>", line 5, in <module>
>     mne.viz.plot_projs_topomap(projs)
>   File "/home/ktavabi/Projects/mne-python/mne/viz/topomap.py", line 264, in plot_projs_topomap
>     % proj['desc'])
> RuntimeError: Cannot find a proper layout for projection Average EEG reference

Still think it's a bug?

larsoner:
Doing it with the average reference is still weird, but I suspect you'll have the same problem if you do `compute_proj_exg(..., n_eeg=1)` so yeah it's still a bug. If you can do `evoked.pick_types(meg=False, eeg=True).plot_topomap()`, you should be able to plot an EEG projection topomap.

kambysese:
Good point. That snippet is actually from [here](https://martinos.org/mne/dev/auto_tutorials/plot_artifacts_correction_ssp.html#sphx-glr-auto-tutorials-plot-artifacts-correction-ssp-py), and `compute_proj_exg(..., n_eeg=1) makes no difference. 

> If you can do evoked.pick_types(meg=False, eeg=True).plot_topomap(), you should be able to plot an EEG projection topomap.

I'll try this in the PR tomorrow

kambysese:
I am sorry was this 

> If you can do evoked.pick_types(meg=False, eeg=True).plot_topomap()

a suggestion for a workaround in this PR, or should I just put this on hold until [bug fix](https://github.com/mne-tools/mne-python/issues/4612)?

larsoner:
Let's wait for the MNE PR to land then this can be done

We should also add the projection topographies to the report for each subject

kambysese:
This works for me as expected. LMKWYT. It would be nice to have this and #184 merged ASAP for my next analysis batch.

mnefun/_mnefun.py
```diff
@@ -2011,6 +2018,9 @@ def do_preprocessing_combined(p, subjects, run_indices):
                 projs.extend(pr)
             else:
                 warnings.warn('Only %d EOG events!' % eog_events.shape[0])
+            if p.plot_pca:
+                h = plot_projs_topomap(pr, info=raw.info, show=False)
+                h.savefig(eog_proj[:-4] + '.png', format='png', dpi=120)
```
larsoner: IIRC there are also empty-room projectors, can you check?

kambysese:
IDK how I keep missing the mnefun notifications.

> WDYT about adding these figures to the report for each subject?

+1 I was thinking in a separate PR, no?

larsoner:
> I was thinking in a separate PR, no?

It's easy to forget to do this so open an issue if you're rather do it later

mnefun/_mnefun.py
kambysese: There are erm projectors but not considered here. Do you want SSP topomaps for them?

kambysese: Done

mnefun/_mnefun.py
```diff
@@ -401,6 +404,7 @@ def __init__(self, tmin=None, tmax=None, t_adjust=0, bmin=-0.2, bmax=0.0,
         self.must_match = []
         self.on_missing = 'error'  # for epochs
         self.subject_run_indices = None
+        self.plot_pca = False
```
larsoner: Since this doesn't change the existing outputs (just adds some image files), actually we might as well set this to `True` I think. Okay with you?

Last comment I swear :)

kambysese: And leave `params.plot_pca = True` as is in `analysis.py`?

larsoner: Either remove it or add a comment that it's the default but is just there for informative purposes

larsoner:
Thanks @kambysese

PR #190, opened 2017-10-24
FIX: Fix for Spyder

larsoner:
Give this a shot @mdclarke

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -14,6 +14,7 @@ import subprocess
 import glob
 from collections import Counter
 import time
+import sys
 
 import numpy as np
 import scipy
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -685,7 +687,7 @@ def fetch_raw_files(p, subjects, run_indices):
             cmd += ['--include', op.basename(fname)]
         remote_loc = '%s:%s' % (p.acq_ssh, op.join(remote_dir, ''))
         cmd += ['--exclude', '*', remote_loc, op.join(raw_dir, '')]
-        run_subprocess(cmd)
+        run_subprocess(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         # move files to root raw_dir
```
ktavabi: [Python pages](https://docs.python.org/2/library/subprocess.html#using-the-subprocess-module) recommends 

> Do not use stdout=PIPE or stderr=PIPE with this function as that can deadlock based on the child process output volume. Use Popen with the communicate() method when you need pipes.

You thinking this use case is simple enough to ignore this?

larsoner: I think it should be okay

larsoner:
@mdclarke reported that this seems to work so I'll merge, but let me know if there are other problems

ktavabi:
Also resolves `pykernel` issue with `subprocess` in pycharm. Changes LGTM aside from comment/question about stdout=PIPE or stderr=PIPE usage.

PR #192, opened 2017-10-24
Fix fwd & inverse compatibility issues.

### LINKED ISSUE #191: read_forward_solution getting deprecated arg
@larsoner `read_forward_solution()` getting deprecated arg `surf_ori`. Will try fixing with `convert_forward_solution()`

ktavabi:
@larsoner please review and confirm loose/free/fixed source ori parameters. Fixes #191

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1623,9 +1623,9 @@ def gen_inverses(p, subjects, run_indices):
             fwd_name = op.join(fwd_dir, s_name + p.inv_tag + '-fwd.fif')
             fwd = read_forward_solution(fwd_name)
             fwd = convert_forward_solution(fwd, surf_ori=True)
-            looses = [0]
+            looses = [1]
             tags = [p.inv_free_tag]
-            fixeds = [True]
+            fixeds = [False]
             depths = [0.8]
             if fwd['src'][0]['type'] == 'surf':
                 looses += [0, 0.2]
@@ -1643,7 +1643,7 @@ def gen_inverses(p, subjects, run_indices):
                     if l < 1:
                         use_cps = True
                     else:
-                        use_cps = False
+                        use_cps = None
                     inv_name = op.join(inv_dir, temp_name + f + s + '-inv.fif')
                     inv = make_inverse_operator(raw.info, fwd_restricted, cov,
                                                 loose=l, depth=d, fixed=x,
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -26,7 +26,7 @@ import mne
 from mne import (compute_proj_raw, make_fixed_length_events, Epochs,
                  find_events, read_events, write_events, concatenate_events,
                  read_cov, compute_covariance, write_cov,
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -1639,9 +1640,14 @@ def gen_inverses(p, subjects, run_indices):
             for f, m, e in zip(out_flags, meg_bools, eeg_bools):
                 fwd_restricted = pick_types_forward(fwd, meg=m, eeg=e)
                 for l, s, x, d in zip(looses, tags, fixeds, depths):
+                    if l < 1:
+                        use_cps = True
+                    else:
+                        use_cps = False
```
larsoner: we should always use CPS

mnefun/_mnefun.py
```diff
@@ -1621,13 +1621,14 @@ def gen_inverses(p, subjects, run_indices):
...
             temp_name = s_name + ('-%d' % p.lp_cut) + p.inv_tag
             fwd_name = op.join(fwd_dir, s_name + p.inv_tag + '-fwd.fif')
-            fwd = read_forward_solution(fwd_name, surf_ori=True)
-            looses = [None]
+            fwd = read_forward_solution(fwd_name)
+            fwd = convert_forward_solution(fwd, surf_ori=True)
+            looses = [0]
             tags = [p.inv_free_tag]
-            fixeds = [False]
+            fixeds = [True]
```
larsoner: False

mnefun/_mnefun.py
```diff
@@ -1621,13 +1621,14 @@ def gen_inverses(p, subjects, run_indices):
             s_name = safe_inserter(name, subj)
             temp_name = s_name + ('-%d' % p.lp_cut) + p.inv_tag
             fwd_name = op.join(fwd_dir, s_name + p.inv_tag + '-fwd.fif')
-            fwd = read_forward_solution(fwd_name, surf_ori=True)
-            looses = [None]
+            fwd = read_forward_solution(fwd_name)
+            fwd = convert_forward_solution(fwd, surf_ori=True)
+            looses = [0]
```
larsoner: [1]

mnefun/_mnefun.py
ktavabi: So `else use_cps = None` for the fixed and free ori cases, correct?

larsoner: I would just always set `use_cps=True`

ktavabi: The [docstr](https://martinos.org/mne/dev/generated/mne.minimum_norm.make_inverse_operator.html) for `make_inverse_operator` says cps is...

>Only used when converting to surface orientation (i.e., for surface source spaces and loose < 1).

Won't it balk with `loose = 1`?

larsoner:
Does it work okay with `use_cps=True` always on?

ktavabi:
> Does it work okay with use_cps=True always on?

Looks that way. It's not balking in the pipeline for funloc!

larsoner:
Thanks for taking care of these!

PR #194, opened 2017-10-25
fix additional subprocess io streams

### LINKED ISSUE #193: subprocess io stream broken in hp estimation routine
Missed a couple of broken `run_subprocess` instances in last fix. 

cc @mdclarke

ktavabi:
@mdclarke try this.

Closes #193.


mnefun/_mnefun.py
```diff
@@ -838,7 +838,7 @@ def run_sss(p, subjects, run_indices):
             print('-' * len(s))
             print(s)
             print('-' * len(s))
-            run_subprocess(cmd, stdout=sys.stdout, stderr=sys.stderr)
+            run_subprocess(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
larsoner: leaving this one blank was intentional so that you can see the output as it goes (plus people don't use it much anymore so if they do, they can do it from the terminal to have it work I think)

ktavabi:
Tested with funloc data.

larsoner:
Thanks @ktavabi

PR #196, opened 2017-10-30
WIP fix make volumetric source space for sphere

ktavabi:
@larsoner couple questions with this one:

1. Wasn't sure about what you discussed with @mdclarke wrt to `pos` argument for `setup_volume_source_space` but I went with `pos=7.0` rather than a dictionary. If you know what `pos[‘rr’]` and `pos[‘nn’]` should look like then I can change that.

2. LMK how you want to handle trans for volumetric sourcespace. `make_forward_solution` seems to expect `type` mne.transforms.Transform class and not a dict. This seems to contradict [docstring](https://martinos.org/mne/stable/generated/mne.make_forward_solution.html)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,8 +40,7 @@ nosetests.xml
 .idea/
 
 # subject files
-examples/funloc/subj*/**/
-examples/funloc/subj_*/**/*.html
+examples/funloc/subj_0*/
 examples/funloc/run_sss.sh
 examples/funloc/.ipynb_checkpoints/
 
@@ -52,3 +51,5 @@ examples/funloc/.ipynb_checkpoints/
 env/
 *.whl
 .cache/
+cachedir/
+examples/funloc/cachedir/
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1689,12 +1689,7 @@ def gen_forwards(p, subjects, structurals, run_indices):
             sphere *= 1000.  # to mm
             src = setup_volume_source_space(subject=subj, pos=7.0, sphere=sphere,
                                             mindist=1.)
-            trans = {'from': FIFF.FIFFV_COORD_HEAD,
-                     'to': FIFF.FIFFV_COORD_MRI,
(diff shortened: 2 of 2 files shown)


larsoner:
I don't think we need to write the trans file. Why add that?

`mne.transforms.Transform` is a subclass of `dict` so while technically it's correct, that doc should probably be updated

ktavabi:
> I don't think we need to write the trans file. Why add that?
IDK. Posterity? I assume I can use `Transform` subclass then. 

> doc should probably be updated
I'll start PR

larsoner:
> I assume I can use Transform subclass then.

Yes you can. But I wouldn't bother, in this case you can directly use `None`:

https://mne-tools.github.io/stable/generated/mne.make_forward_solution.html#mne.make_forward_solution

ktavabi:
If `trans=None` then we shouldn't need this either, correct?

```
trans = {'from': FIFF.FIFFV_COORD_HEAD,
                     'to': FIFF.FIFFV_COORD_MRI,
                     'trans': np.eye(4)}
```

larsoner:
right

.gitignore
```diff
@@ -40,8 +40,8 @@ nosetests.xml
 .idea/
 
 # subject files
-examples/funloc/subj*/**/
-examples/funloc/subj_*/**/*.html
+/examples/funloc/subj01/
+/examples/funloc/subj02/
```
larsoner: these changes look incorrect -- why add the leading slash, and why change to `subj01` when the name is `subj_01`?

ktavabi: Sorry that's a typo from my end. The leading slash was inserted by pycharm.

larsoner:
LGTM, good to go from your end @ktavabi ?

ktavabi:
Worked with funloc data. @mdclarke latest commit work for you?

mdclarke:
Yep! all good. Thanks!

larsoner:
Thanks @ktavabi

PR #197, opened 2017-11-21
WIP plot bem+src space alignment

ktavabi:
@larsoner WDYT? If yes, then would it makes more sense to combine the three mayavi scenes into a single figure to insert into Report?

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -21,6 +21,7 @@ import scipy
 from scipy import io as spio
 import matplotlib.pyplot as plt
 from numpy.testing import assert_allclose
+from mayavi import mlab
 
 import mne
 from mne import (compute_proj_raw, make_fixed_length_events, Epochs,
@@ -253,9 +254,6 @@ class Params(Frozen):
     plot_pca : bool
         If set to True generate selected PCA component topographies and save
         figures to disk in pca_fif folder.
-    plot_sphere_bem: bool
-        If using spherical BEM conductor model and set to True then plot
-        data-source space alignment and save figure to disk in inverse folder.
 
     Returns
     -------
@@ -416,7 +414,6 @@ class Params(Frozen):
         self.subject_run_indices = None
         self.autoreject_thresholds = False
         self.plot_pca = True
-        self.plot_sphere_bem = False
         self.freeze()
 
     @property
@@ -1693,23 +1690,7 @@ def gen_forwards(p, subjects, structurals, run_indices):
             src = setup_volume_source_space(subject=subj, sphere=bem,
                                             pos=10.)
             trans = None
-            bem_type = 'spherical model'
-            if p.plot_sphere_bem:
-                fg = mne.viz.plot_alignment(info, meg=['helmet', 'sensors'], bem=bem,
-                                            src=src, dig=True,
-                                            surfaces=['brain', 'inner_skull',
-                                                      'outer_skull',
-                                                      'outer_skin'])
-                fg.scene._toggle_axes()
-                for nm, ix in zip(['lt', 'rt', 'nz'], [90, 180, 90]):
-                    fg.scene.camera.pitch(0)
-                    fg.scene.camera.azimuth(ix)
-                    fg.scene.reset_zoom()
-                    fname = op.join(p.work_dir, subj, p.inverse_dir,
-                                op.basename(raw_fname)[:-4] +
-                                '_%s_spherical-bem.png' % nm)
-                    fg.scene.save_png(fname)
-                fg.scene.close()
+            bem_type = 'spherical-BEM'
         else:
             trans = op.join(p.work_dir, subj, p.trans_dir, subj + '-trans.fif')
             if not op.isfile(trans):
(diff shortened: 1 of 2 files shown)


larsoner:
We should probably always call `plot_alignment` with whatever source space and conductor model the user has chasen. No need to even have a binary for spherical source spaces or conductor models, we should just add a `plot_alignment` image of whatever is used.

I would plot it in MEG coordinate frame, and set the view to left, center, right, look ~2/3 of the way down the page here for code and output to get started:

https://larsoner.github.io/mne-biomag-group-demo/auto_examples/single_subject_analysis/plot_analysis_01.html#sphx-glr-auto-examples-single-subject-analysis-plot-analysis-01-py

ktavabi:
> and set the view to left, center, right
Not exactly sure what you mean here. This is what I have so far tested with funloc datasets.

image

larsoner:
Yeah that's what I mean

mnefun/_mnefun.py
```diff
@@ -21,6 +21,7 @@
 from scipy import io as spio
 import matplotlib.pyplot as plt
 from numpy.testing import assert_allclose
+from mayavi import mlab
```
larsoner: nest this, don't make it a hard dependency

mnefun/_mnefun.py
```diff
@@ -1683,14 +1684,13 @@ def gen_forwards(p, subjects, structurals, run_indices):
...
             # create source space
-            sphere = np.concatenate((bem['r0'], [bem['layers'][0]['rad']]))
-            sphere *= 1000.  # to mm
-            src = setup_volume_source_space(subject=subj, pos=7.0, sphere=sphere,
-                                            mindist=1.)
+            src = setup_volume_source_space(subject=subj, sphere=bem,
+                                            pos=10.)
             trans = None
-            bem_type = 'spherical model'
+            bem_type = 'spherical-BEM'
```
larsoner: `spherical conductor model` if you want but `spherical BEM` is a misnomer

mnefun/_mnefun.py
```diff
@@ -1683,14 +1684,13 @@ def gen_forwards(p, subjects, structurals, run_indices):
         subjects_dir = get_config('SUBJECTS_DIR')
         if structurals[si] is None:  # spherical case
             # create spherical BEM
-            bem = make_sphere_model('auto', 'auto', info, verbose=False)
+            bem = make_sphere_model(info=info, r0='auto',
+                                    head_radius='auto')
```
larsoner: why this change? The `verbose=False` is nice to avoid fitting stuff. Feel free to emit a `print(...)` if you want a useful output.

mnefun/_mnefun.py
```diff
@@ -1683,14 +1684,13 @@ def gen_forwards(p, subjects, structurals, run_indices):
...
             # create source space
-            sphere = np.concatenate((bem['r0'], [bem['layers'][0]['rad']]))
-            sphere *= 1000.  # to mm
-            src = setup_volume_source_space(subject=subj, pos=7.0, sphere=sphere,
-                                            mindist=1.)
+            src = setup_volume_source_space(subject=subj, sphere=bem,
+                                            pos=10.)
```
larsoner: `pos` should probably default to `7.` but be configurable

larsoner: (same with `mindist`, the old default of `1.` is probably fine)

mnefun/_mnefun.py
```diff
@@ -1725,6 +1725,36 @@ def gen_forwards(p, subjects, structurals, run_indices):
             fwd = make_forward_solution(
                 info, trans, src, bem, n_jobs=p.n_jobs, mindist=p.fwd_mindist)
             write_forward_solution(fwd_name, fwd, overwrite=True)
+        # plot BEM & source space alignment
+        plt.rcParams['axes.titlesize'] = 'medium'
+        plt.rcParams['axes.titleweight'] = 'normal'
```
larsoner: This will make permanent changes (bad), use:

https://matplotlib.org/devdocs/api/_as_gen/matplotlib.pyplot.rc_context.html

Or just tweak the axes or object later (probably easiest)

mnefun/_mnefun.py
```diff
@@ -1725,6 +1725,36 @@ def gen_forwards(p, subjects, structurals, run_indices):
...
+            axes[ai].imshow(view, interpolation='bicubic')
+            axes[ai].set_title(tv, fontdict=fontdict)
+        mlab.close(aln)
+        fig.subplots_adjust(left=0, right=1, top=1, bottom=0, wspace=0.05,
+                            hspace=0)
+        fname = op.join(p.work_dir, subj, p.inverse_dir,
+                        op.basename(raw_fname)[:-4] +
+                        '_%s.png' % bem_type)
+        fig.savefig(fname, dpi=150, facecolor=fig.get_facecolor(),
+                    edgecolor='none')
```
larsoner: I don't really like the proliferation of all of these images. Why don't we just add all of them to the report? Is the report generation to slow or not useful enough?

ktavabi: Will address comments. 

I am also not fond of the proliferation of images. Honestly, It's just me trying to get something done promptly and have just been holding off on messing with the report. IDK how useful others find it; I've asked the MEG team what if anything should be included in the report (or better yet open Github issues) wrt QC but crickets. 

I can start a PR to add:
1. cHPI SNR
2. PCA topomaps
3. Alignment
???

larsoner: That would be a great start, feel free to get one or more of those in a PR and we can do follow-up PRs for others we think of

PR #198, opened 2017-11-22
FIX: trans_to, subjects_dir, extract_expyfun_events

### LINKED ISSUE #195: extract_expyfun_events fails if STI001-8 is not there
We can use STI101 instead in this case. cc @mdclarke

larsoner:
This PR:

1. Adds a fourth argument to the `trans_to` tuple, which is a forward(+)/backward(-) head rotation, e.g. `(0., 0., 0.04, -30)` will rotate the head backward 30 degrees.
2. Allows setting `subjects_dir` that will be used to find the structurals.
3. Fixes processing when `STI101` is the only trigger channel available. @mdclarke can you confirm this fixes #195 for you?

Closes #195.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1709,8 +1709,9 @@ def gen_forwards(p, subjects, structurals, run_indices):
         raw_fname = get_raw_fnames(p, subj, 'sss', False, False,
                                    run_indices[si])[0]
         info = read_info(raw_fname)
+        struc = structurals[si]
 
-        if structurals[si] is None:  # spherical case
+        if struc is None:  # spherical case
             # create spherical BEM
             bem = make_sphere_model('auto', 'auto', info, verbose=False)
             # create source space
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -1699,7 +1728,7 @@ def gen_forwards(p, subjects, structurals, run_indices):
                 if not op.isfile(trans):
                     raise IOError('Unable to find head<->MRI trans file')
             src_space_file = op.join(subjects_dir, structurals[si], 'bem',
-                                     structurals[si] + '-oct-6-src.fif')
+                                     structurals[si] + '-oct6-src.fif')
```
larsoner: Argh, need to revert/adapt this

ktavabi: I am pretty sure we use `struc = structurals[si]` in `gen_forwards`. Any reason why we can't substitute thereafter?
Otherwise LGTM

mdclarke: Works to fix https://github.com/LABSN/mnefun/issues/195 for me

larsoner:
Done @ktavabi, feel free to look again if you have time

mnefun/_mnefun.py
```diff
@@ -1727,16 +1728,19 @@ def gen_forwards(p, subjects, structurals, run_indices):
                                 subj + '-trans_head2mri.txt')
                 if not op.isfile(trans):
                     raise IOError('Unable to find head<->MRI trans file')
-            src_space_file = op.join(subjects_dir, structurals[si], 'bem',
-                                     structurals[si] + '-oct6-src.fif')
-            if not op.isfile(src_space_file):
+            for mid in ('oct6', 'oct-6'):
+                src_space_file = op.join(subjects_dir, struc, 'bem',
+                                         '%s-%s-src.fif' % (struc, mid))
```
ktavabi: Isn't just cleaner if `src_space_files ` are named as expected i.e., `np-src`, renaming can be done with a simple bash command. Or is this addressing some other issue IDK about?

ktavabi:
Other than comment LGTM. Thanks @larsoner!

mnefun/_mnefun.py
larsoner: Yeah it would be cleaner. I indeed added this to address a corner case where @mdclarke had a morphed structural that used the `npsrc` instead of `np-src` style naming. I structured this to continue to prefer our `np-src`-style naming, though, so I think we should be okay

PR #199, opened 2017-11-22
MRG: move images to mne-report

### LINKED ISSUE #186: ENH: Add SSP topomaps to subj Report
Add artifactual SSP projection topography maps to mne-python report file.

### LINKED ISSUE #201: Coil dist plot has no show arg
@larsoner just noticed that `plot_good_coils` has no `show` arg. Perhaps adding it in #199?

### LINKED ISSUE #202: BUG raw files with orig extension should be filtered
`get_raw_fnames(p, subj, 'raw', erm=False, add_splits=True, indices=run_indices[si])` is returning raw files with .orig extension for reporting in #199. @larsoner I can't think of a reason these files should not be filtered out?  In my case files for .orig are being sent to KASGA for headpos estimation and causing remote errors. 

Perhaps modifying `_regex_convert(f)` to return `.*%s-?[0-9]*.fif(?!.orig) % op.basename(f)[:-4]`?

ktavabi:
Replaces #197 and fixes #186.  This PR will collate following figures in subjects mne-report.

- [x] src space alignment
- [x] PCA topomaps
- [x] cHPI SNR
- [x] head movement traces
- [x] diagnostic plots (e.g., good coils, head movement, etc.) should go at the very top
- [x] if using BEM rather than sphere model call add_bem_to_section manually.
- [x] check to see if files exist before trying to use them
- [x] set the Report regex to include only some specific filenames
- [x] parse_folder explicitly with render_bem=False
- [x] handle case where `p.trans_to == 'median'`

Closes #201.
Closes #202.


larsoner:
@ktavabi are you planning on working on this this week? If not, okay if I take over?

I'm also thinking #155 can be rolled in with this.

ktavabi:
@larsoner I'm currently trying to streamline PCA plots. #155 hasn't been on
my radar but I can revisit.

--
Kambiz Tavabi
*Sent from a magical device that lives in my pocket. please excuse brevitas
& typos.*

ktavabi:
@larsoner here is what I have so far. The reports are beginning to look good but could do with a bit of reordering and section nomenclature. LMKWYT. 
Also, this branch has diverged sig from master; how do I resolve conflicts in this case? 
1. Rebase `revamp-mne-report` onto `master`
or 
2. Merge `master` into `revamp-mne-report` (I think it's this one, right?)

larsoner:
Whichever will cause you less consternation. Maintainers can squash+merge on the GitHub interface so maintaining a clean history with rebasing is less critical.

larsoner:
On the first run I get:
```
IOError: [Errno 2] No such file or directory: '/home/larsoner/Documents/python/mnefun/examples/funloc/subj_01/sss_pca_fif/subj_01_funloc_allclean_fil55_raw_sss.fif'
```
So you should probably check to see if files exist before trying to use them.

I also get a lot of these:
```
/home/larsoner/.local/lib/python2.7/site-packages/joblib/parallel.py:92: RuntimeWarning: Failed to process file /home/larsoner/Documents/python/mnefun/examples/funloc/subj_01/raw_fif/subj_01_erm_raw.fif:
"This file contains raw Internal Active Shielding data. It may be distorted. Elekta recommends it be run through MaxFilter to produce reliable results. Consider closing the file and running MaxFilter on the data. Use allow_maxshield=True if you are sure you want to load the data despite this warning."
```
Partly this is a `Report` bug that it should handle MaxShield data (will fix in MNE), but also we probably want to set the `Report` regex to include only some specific filenames (rather than the default, which tries to parse all `.fif` files).

larsoner:
It also takes a looooong time to generate the report. Looks like it's not a regex but a list, and looking at the code I see you already set it. We should figure out what takes so long in the generation step and fix it. Does it take a while for you? Maybe I have some `joblib` bug or something.

We probably also want to do the `parse_folder` explicitly with `render_bem=False`, and then if we are actually using a BEM rather than a sphere model call `add_bem_to_section` manually.

larsoner:
For some reason for me it hangs on the `report.parse_folder` call due to `n_jobs`, maybe we should always set it to `1` for now.

The `plot_alignment` looks a little bit small, maybe it's worth making it vertically stacked and much larger, instead?

larsoner:
One last thing -- I think that the diagnostic plots (e.g., good coils, head movement, etc.) should go at the very top

ktavabi:
> Partly this is a Report bug that it should handle MaxShield data (will fix in MNE)

Will you open issue?

>It also takes a looooong time to generate the report. We should figure out what takes so long in the generation step and fix it. Does it take a while for you? Maybe I have some joblib bug or something.

I am getting very different subjective timing results on my WS (Ubuntu 16.04) vs. MBP. I was considering `timeit` but not sure where to do it. I ran `report` using DEBUG verbosity and the rate limiters seem to be the evokeds! And BEM. But that doesn't rule out a joblib/otherwise bug.

>We probably also want to do the parse_folder explicitly with render_bem=False, and then if we are actually using a BEM rather than a sphere model call add_bem_to_section manually.

I have to look into this I'm not sure WYM exactly. 

>One last thing -- I think that the diagnostic plots (e.g., good coils, head movement, etc.) should go at the very top

I was unable to figure out how to re-order sections in `report`. Either due to the lack of python prowess on my behalf or flexibility in `Report`

I hate regex :sob:

larsoner:
> Will you open issue?

I already made a PR to fix it

> I was unable to figure out how to re-order sections in report

I would try adding the sections *first*, then call `parse_folder` *second* (or multiple times with different args). That hopefully will define the order.

> I hate regex

Don't worry about regex, we can pass exact filenames.

ktavabi:
>Whichever will cause you less consternation

For that uncomfortable feeling of consternation with git. 
I GitKraken.
By using GitKraken for all your pushing and pulling needs. Now you too can enjoy life without consternation. I did.

This ad is brought to you by GitKraken.

mnefun/_mnefun.py
```diff
@@ -2411,21 +2387,104 @@ def gen_html_report(p, subjects, structurals, run_indices=None,
...
+            n_jobs = 1
+        report.parse_folder(data_path=path, mri_decim=10, n_jobs=n_jobs,
+                            pattern=patterns, sort_sections=True)
+        # Add HP coils SNR
+        fig = plot_good_coils(read_raw_fif(raw_fif_fn, allow_maxshield='yes'))
+        report.add_figs_to_section(fig, captions='cHPI coils',
+                                   section='cHPI')
+        # Add head movement traces
+        pos = _headpos(p, raw_fif_fn)
+        fig = plot_head_positions(pos=pos, show=False)
```
larsoner: We should modify this figure to show the median or time-weighted head position values using `ax.hlines`. I can do this in a follow-up PR, though, after this one is merged.

.ropeproject/config.py
```diff
@@ -0,0 +1,112 @@
+# The default ``config.py``
+# flake8: noqa
+
+
+def set_prefs(prefs):
```
larsoner: is this somehow generally useful? maybe instead we should add `.ropeproject` to `.gitignore`

larsoner:
@ktavabi I have some time to look into this today, can I push to your branch?

ktavabi:
Yes, please do.

--
Kambiz Tavabi
*Sent from a magical device that lives in my pocket. please excuse brevitas
& typos.*

.ropeproject/config.py
ktavabi: I have no idea what this is! :confused:  Thinking related to ATOM I've added to ignore.

ktavabi:
Hope I am not too late. I've pushed what I am going to do for now. Please take over. Couple of things:
1. I am still getting subjectively different performance on Ubuntu WS (python 2.7 venv) vs. MBP (python 3.6 venv).  Currently, all I am doing is plotting PSD and cHPI SNR (which is kind of slow), and I can't figure out how to even use `timeit.timeit` to see what's taking so long.
2. `Report.parse_folder()` seems to override existing report with custom sections i.e., replacing/displacing diagnostic plots on top.

larsoner:
Okay I'll take over

mnefun/_mnefun.py
```diff
@@ -1122,6 +1110,33 @@ def run_sss_locally(p, subjects, run_indices):
...
 
 
+def _load_trans_to(p, subj, run_indices, raw=None):
+    if isinstance(p.trans_to, string_types):
+        if p.trans_to == 'median':
+            trans_to = op.join(p.work_dir, subj, p.raw_dir,
+                               subj + '_median_pos.fif')
+            if not op.isfile(trans_to):
+                calc_median_hp(p, subj, trans_to, run_indices)
+        trans_to = mne.read_trans(trans_to)
```
ktavabi: Currently fails in case where `trans_to == 'twa'`

mnefun/_mnefun.py
```diff
@@ -1122,6 +1110,33 @@ def run_sss_locally(p, subjects, run_indices):
...
+            t[:3, 3] = trans_to[:3]
+            theta = np.deg2rad(trans_to[3])
+            t[1:3, 1:3] = [[np.cos(theta), -np.sin(theta)],
+                           [np.sin(theta), np.cos(theta)]]
+            trans_to = mne.Transform('meg', 'head', t)
+        elif trans_to.shape != (3,):
+            raise ValueError('trans_to must have 3 or 4 elements, '
+                             'got shape %s' % (trans_to.shape,))
+    if trans_to is not None:
+        trans_to = mne.transforms._ensure_trans(trans_to, 'meg', 'head')
```
ktavabi: For some reason `_ensure_trans` is returning a 
>ValueError: trans must be a Transform between meg<->head, got None

when `params.trasn_to = (0, 0, .03)`

mnefun/_mnefun.py
larsoner: it should, #200 is not in `master`

larsoner:
I have no issue with `p.trans_to = 'median'`, it's the default and what I was testing with `analysis_fun.py`

larsoner:
The `params.trans_to = (0., 0., 0.03)` case should be fixed now, feel free to try it.

If you think it looks good, we can merge this, and I can rebase #200.

mnefun/_mnefun.py
ktavabi: I figured it will once that is merged. Just making sure.

ktavabi:
WDYT of #202? should we fix now or merge this first?

larsoner:
Already fixed in this PR:

https://github.com/LABSN/mnefun/pull/199/commits/a96cdce1dfb083c8a7143c9d0e45880b8c9086ed

ktavabi:
Ugh...you always do it in a fraction of the code.

PR #205, opened 2017-12-20
WIP: Add annotations

larsoner:
Todo before merge:

- [x] Get MNE Annotation I/O PR merged.
- [x] Save to disk, load from disk as necessary
- [x] Incorporate in `mnefun` per run
- [x] Adjust time-weighted average using the removed sections
- [ ] Test with problematic data (@ktavabi)
- ~~[ ] Check for unreasonable head positions by fitting a sphere and checking for sensor collisions~~

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -2887,3 +2887,90 @@ def _spherical_conductor(info, subject, pos):
     src = setup_volume_source_space(subject=subject, sphere=bem,
                                     pos=pos, mindist=1.)
     return bem, src, None
+
+
+def annotate_head_pos(raw, head_pos, rotation_limit=45, translation_limit=0.1,
+                      coil_dists=True, t_step=0.1, t_window=0.1,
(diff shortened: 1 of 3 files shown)


larsoner:
I'm going to punt on the checking for unreasonable positions using a sphere for now, it can be done separately later. Let me know here or on Slack how it goes @ktavabi

mnefun/_mnefun.py
```diff
@@ -2456,16 +2478,38 @@ def _prebad(p, subj):
     return prebad_file
 
 
-def _headpos(p, raw_fname):
-    """Helper for locating head position estimation file"""
+def _head_pos_annot(p, raw_fname, prefix='  '):
```
ktavabi: this refactor breaks `_report.py`

ktavabi:
mne `plot_head_positions` is throwing:
>ValueError: setting an array element with a sequence.
I think b/c `trans_to` as `{Transform}` dict type is not a valid kwarg.

PR #209, opened 2018-02-08
WIP: Better good check

larsoner:
@ktavabi I have made some improvements to the cHPI fitting and annotations that might help. When you get a chance, can you try this on the subject who ended up with ~6 usable trials after adding annotations:

1. make a note of how many epochs you currently have (6?)
2. update to latest `mne` (so that it includes some recent cHPI fixes)
3. switch to this PR on `mnefun`
4. delete the `raw_fif/*-annot.fif` and `raw_fit/*-counts.h5` files, if they exist
5. delete the `raw_fif/*twa_pos.fif` files
6. rerun the subject's steps from Maxfilter onward (`do_sss`, `fetch_sss`, `do_ch_fix`, `gen_ssp`, `apply_ssp`, `write_epochs`), noting what percentage of trials are dropped for what reason
7. note how many epochs you end up with
8. Repeat steps 4-7 after setting `p.coil_dist_limit = 0.01` and see if you get more usable trials
9. Repeat steps 4-7 after setting `p.coil_t_window = 'auto'` and see if you get more usable trials.
10. Repeat steps 4-7 after setting `p.coil_bad_count_duration_limit = 1.` or so and see if you get more usable trials. 

Todo:

- [ ] modify coil distance constraint for coil fitting to ensure that the coils stay far enough away from the sensors

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -28,6 +28,7 @@ from mne import (
     setup_volume_source_space, pick_info, write_source_spaces,
     read_source_spaces, write_forward_solution, DipoleFixed,
     read_annotations)
+from mne.externals.h5io import read_hdf5, write_hdf5
 
 try:
     from mne import compute_raw_covariance  # up-to-date mne-python
@@ -52,7 +53,8 @@ from mne.minimum_norm import make_inverse_operator
 from mne.label import read_label
 from mne.epochs import combine_event_ids
 from mne.chpi import (filter_chpi, read_head_pos, write_head_pos,
-                      _get_hpi_info)
+                      _get_hpi_info, _get_hpi_initial_fit, _setup_hpi_struct,
+                      _fit_cHPI_amplitudes, _fit_magnetic_dipole)
 from mne.io.proj import _needs_eeg_average_ref_proj
 
 from mne.cov import regularize
@@ -414,20 +416,20 @@ class Params(Frozen):
         self.subjects_dir = None
         self.src_pos = 7.
         self.report_params = dict(
-            coil_snr=True,
+            good_hpi_count=True,
             head_movement=True,
             psd=True,
             ssp_topomaps=True,
             source_alignment=True,
             bem=True,
-            coil_t_step='auto',
             source=None,
             )
         self.rotation_limit = np.inf
         self.translation_limit = np.inf
-        self.coil_dist_limit = np.inf
-        self.coil_t_window = 0.05
+        self.coil_dist_limit = 0.005
+        self.coil_t_window = 0.2  # default is same as MF
         self.coil_t_step_min = 0.01
+        self.annotate_bad_coil_counts = False
         self.freeze()
 
     @property
@@ -754,7 +756,7 @@ def calc_twa_hp(p, subj, out_file, ridx):
     for raw_fname in raw_fnames:
         raw = mne.io.read_raw_fif(raw_fname, allow_maxshield='yes',
                                   verbose='error')
-        hp, annot = _head_pos_annot(p, raw_fname, prefix='          ')
+        hp, annot, _ = _head_pos_annot(p, raw_fname, prefix='          ')
         raw.annotations = annot
         good = np.ones(len(raw.times))
         ts = np.concatenate((hp[:, 0],
@@ -1002,7 +1004,8 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
 
 
 def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22,
-                      prefix='  ', work_dir='~/'):
+                      prefix='  ', work_dir='~/', t_window=None,
+                      t_step_min=None, dist_limit=None):
     """Run Maxfilter remotely and fetch resulting file
 
     Parameters
@@ -1023,6 +1026,10 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22,
         The prefix to use when printing status updates.
     work_dir : str
         Where to store the temporary files.
+    t_window : float | None
+        Time window (sec) to use.
+    dist_limit : float | None
+        Distance limit (m) to use.
     """
     # let's make sure we can actually write where we want
     if not op.isfile(fname_in):
@@ -1037,6 +1044,12 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22,
             fnames_in.append(next_name)
         else:
             break
+    if t_window is not None:
+        opts += ' -hpiwin %d' % (round(1000 * t_window),)
+    if t_step_min is not None:
+        opts += ' -hpistep %d' % (round(1000 * t_step_min),)
+    if dist_limit is not None:
+        opts += ' -hpie %d' % (round(1000 * dist_limit),)
 
     t0 = time.time()
     print('%sOn %s: copying' % (prefix, host), end='')
@@ -1053,7 +1066,7 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='', port=22,
         remote_out = op.join(work_dir, 'temp_%s_raw_quat.fif' % t0)
         remote_hp = op.join(work_dir, 'temp_%s_hp.txt' % t0)
 
-        print(', running -headpos', end='')
+        print(', running -headpos%s' % opts, end='')
         cmd = ['ssh', '-p', str(port), host,
                '/neuro/bin/util/maxfilter -f ' + remote_ins[fi] + ' -o ' +
                remote_out +
@@ -1127,7 +1140,7 @@ def run_sss_locally(p, subjects, run_indices):
             print('      Processing %s ...' % op.basename(r))
 
             # estimate head position for movement compensation
-            head_pos, annot = _head_pos_annot(p, r, prefix='        ')
+            head_pos, annot, _ = _head_pos_annot(p, r, prefix='        ')
             raw.annotations = annot
 
             # get the destination head position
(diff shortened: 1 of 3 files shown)


mnefun/_mnefun.py
```diff
@@ -2503,20 +2501,24 @@ def _head_pos_annot(p, raw_fname, prefix='  '):
...
+                               '  translation_limit=%s\n' % lims[1],
+                               '  coil_dist_limit=%s, t_step=%s, t_window=%s'
+                               % tuple(lims[2:])]))
             annot = annotate_head_pos(
                 raw, head_pos, rotation_limit=lims[0],
                 translation_limit=lims[1],
-                coil_dist_limit=lims[2], prefix='  ' + prefix)
+                coil_dist_limit=lims[2], t_step=lims[3], t_window=lims[4],
+                prefix='  ' + prefix)
             annot.save(annot_fname)
```
ktavabi: Are annotation parameters now mandatory?
>   File ".../mnefun_/process_mismatch.py", line 157, in <module>
>     print_status=True
>   File .../mnefun/mnefun/_mnefun.py", line 627, in do_processing
>     outs[ii] = func(p, subjects, structurals, run_indices)
>   File ".../mnefun/mnefun/_report.py", line 96, in gen_html_report
>     for ri, fname in enumerate(fnames)]
>   File ".../mnefun/mnefun/_report.py", line 96, in <listcomp>
>     for ri, fname in enumerate(fnames)]
>   File ".../mnefun/mnefun/_mnefun.py", line 2522, in _head_pos_annot
>     annot.save(annot_fname)
> AttributeError: 'NoneType' object has no attribute 'save'

larsoner: No, sounds like a bug

ktavabi: I think it's this bit, evaluates `False` by default
```
if not np.isfinite(lims).any():
    annot = None
```

larsoner: That part is fine, if no annotations are requested none should be saved. We should just add a `if annot is not None` before this `.save` line.

ktavabi: You want me to do it?

larsoner: Nah it can be part of this PR

larsoner:
Okay @ktavabi see if that fixes the problem

mnefun/_mnefun.py
ktavabi: Ok. So that I understand, we are annotating by default for `twa` and if limits are defined annotations are saved for downstream use?

ktavabi:
Error is resolved but reports are botched without error
image

larsoner:
What is the value of `p.report_params['coil_t_step']`? It is the only thing that should change the HPI coil SNR plot for now.

mnefun/_mnefun.py
larsoner: >  we are annotating by default for twa and if limits are defined annotations are saved for downstream use?

Not by default for TWA (or any other `trans_to` param). You need to enable annotations explicitly by setting some finite requirement values.

ktavabi:
DYM `params.coil_t_step_min = 1.`? I am getting same result with `funloc` data with just this modification to `analysis.py`
```
params.report_params = dict(
            coil_snr=True,
            head_movement=True,
            psd=False,
            ssp_topomaps=False,
            source_alignment=False,
            bem=False,
            coil_t_step='auto',
            source=None,
            )
```

larsoner:
I don't have time to fix now, hopefully early next week. In the meantime maybe you can have a look, or hopefully it does not prevent you from trying the steps I mentioned above

larsoner:
@mdclarke I think I still need to change the estimation step to use the `t_window` in fitting, right now it might just be for our distance checks

larsoner:
@mdclarke options to set the window size (including `'auto'` to do what I described at the meeting) are now included.

@ktavabi this will hopefully give you the flexibility to test your datasets.

larsoner:
@ktavabi it would still be good to do the steps outlined above, but then also check to see if:

8. setting `params.coil_t_window = 'auto'` helps, and
9. setting `params.coil_dist_limit = 0.01` helps

larsoner:
(I guess step (9) is actually already up there implicitly, I'll just add them to the top comment)

mnefun/_report.py
```diff
@@ -67,15 +67,12 @@ def gen_html_report(p, subjects, structurals, run_indices=None,
             #
             # Head coils
             #
-            section = 'HPI coil SNR'
-            if p.report_params.get('coil_snr', True) and p.movecomp:
+            section = 'Good HPI count'
+            if p.report_params.get('good_hpi_count', True) and p.movecomp:
                 t0 = time.time()
+                _, _, fit_data = _head_pos_annot(p, fname, prefix='      ')
```
ktavabi: > Traceback (most recent call last):
>   File "/home/ktavabi/Projects/bad-baby/mnefun_/process_mismatch.py", line 163, in <module>
>     print_status=True
>   File "/home/ktavabi/Projects/mnefun/mnefun/_mnefun.py", line 629, in do_processing
>     outs[ii] = func(p, subjects, structurals, run_indices)
>   File "/home/ktavabi/Projects/mnefun/mnefun/_report.py", line 73, in gen_html_report
>     _, _, fit_data = _head_pos_annot(p, fname, prefix='      ')
> NameError: name 'fname' is not defined

mnefun/_mnefun.py
```diff
@@ -2492,34 +2503,67 @@ def _head_pos_annot(p, raw_fname, prefix='  '):
     """Locate head position estimation file and do annotations."""
```
ktavabi: I can't tell, will this handle split files?

mnefun/_report.py
ktavabi: @larsoner I am getting around this using `raw._filenames[0]` but IDK if this is a good situation for raw files with splits unless `_head_pos_annot` can deal with them downstream?

mnefun/_mnefun.py
larsoner: I hope so, it is what we use for `maxwell_filter` movement compensation IIRC but I'll have to check

larsoner: Looking at the code, `run_sss_positions` looks for extra files like `-1.fif` etc., runs head pos estimation on them all, loads the results, and concatenates them. So just like loading `_raw.fif` in MNE automatically loads the `-1.fif` files as necessary, our `.pos` files should have all positions in them.

ktavabi:
So it looks like gains with the more granular annotation of data based on cHPI data are mixed:

1. After visual inspection of cHPI coil distance plots clearly some datasets show no improvement (n=5) and I think should be excluded from further analysis. Roughly an equal number of cases (n=8) show improvement or very little difference with optimized approach, although in all cases there is more sparsity in the distribution of good/bad segments. 
2. In 3 cases I saw drops between 7-10 in the number of acceptable trials with the optimized HPI estimation approach
3. I got punished in several cases where >95% of trials were dropped due to segments of BAD_RV or BAD_TV
    i. It doesn't seem that 2x or 3x the default values for `p.rotation_limit` and `p.translation_limit`       improves the situation at all. So resorting to `np.inf` now.

cc @staulu

larsoner:
@ktavabi there are a number of parameters that can be adjusted here, at least:

1. `p.rotation_limit` and `p.translation_limit`
2. `p.coil_dist_limit`
3. `p.coil_t_window`
4. `p.coil_bad_count_duration_limit`

So far you only talk about number (1) in this list. Have you tried the others? For example I remember estting `dist_limit = 0.01` may have made one file more usable. Your comment about increased sparsity of bad segments makes me think that `coil_bad_count_duration` might matter. In the top comment I suggested using a systematic approach of re-running all subjects while tweaking these parameters (steps 8-10 in that comment), have you had a chance to / will you have a chance to do this? Without systematic variations it will be difficult to tell what actually helps. If you can send me the analysis script you are using, I can probably script it and run it if it would be too burdensome at your end.

mnefun/_report.py
```diff
@@ -62,23 +62,30 @@ def gen_html_report(p, subjects, structurals, run_indices=None,
                                     run_indices[si])
         has_pca = all(op.isfile(fname) for fname in pca_fnames)
 
+        # whitening and source localization
+        cov_name = op.join(p.work_dir, subj, p.cov_dir,
+                           safe_inserter(cov, subj))
```
ktavabi: _TypeError: argument of type 'bool' is not iterable_
 >  File "..._report.py", line 67, in gen_html_report
>     safe_inserter(cov, subj))
>   File "..._paths.py", line 29, in safe_inserter
>     if '%s' in string:

larsoner: Should be fixed in latest commit but I can't test locally right now (until Monday)

ktavabi:
- [x] HP estimation with p.coil_dist_limit = 0.01 
- [x] HP estimation with p.coil_t_window = 'auto' 
- [x] HP estimation with p.coil_bad_count_duration_limit = 1.
- [ ] HP estimation with p.coil_dist_limit=0.005

mnefun/_report.py
ktavabi: Thanks.

mnefun/_mnefun.py
```diff
@@ -1129,7 +1140,7 @@ def run_sss_locally(p, subjects, run_indices):
             print('      Processing %s ...' % op.basename(r))
 
             # estimate head position for movement compensation
-            head_pos, annot = _head_pos_annot(p, r, prefix='        ')
+            head_pos, annot, _ = _head_pos_annot(p, r, prefix='        ')
```
ktavabi: I don't think we should end up here if `p.movecomp = None`

>   File ".../nbwr/mnefun_/process.py", line 114, in <module>
>     print_status=True
>   File ".../mnefun/_mnefun.py", line 638, in do_processing
>     outs[ii] = func(p, subjects, run_indices)
>   File ".../mnefun/_mnefun.py", line 879, in run_sss
>     run_sss_locally(p, subjects, run_indices)
>   File ".../mnefun/_mnefun.py", line 1143, in run_sss_locally
>     head_pos, annot, _ = _head_pos_annot(p, r, prefix='        ')
> TypeError: 'NoneType' object is not iterable

larsoner: Indeed, fixed

mnefun/_mnefun.py
```diff
@@ -3019,61 +3030,51 @@ def _spherical_conductor(info, subject, pos):
...
-    t_window : float
-        The time window for coil calculations.
+    fit_t : ndarray
+        Fit times.
+    counts : ndarray
+        Coil counts.
+    prefix : str
+        The prefix for printing.
     coil_bad_count_duration_limit : float
         The lower limit for bad coil counts to remove segments of data.
```
ktavabi: Need doc string for `coil_t_window`, and default unit for `coil_bad_count_duration_limit`

larsoner:
@ktavabi I'm going to merge this and we can iterate further if we find problems.

ktavabi:
I was just about to say that. I am +1 for merge.

PR #214, opened 2018-03-13
WIP add function for writing out group source data

ktavabi:
@larsoner no where close to done but WDYT of the overall gist

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -3036,3 +3036,87 @@ def annotate_head_pos(raw, head_pos, rotation_limit=45, translation_limit=0.1,
 
     # Annotate on distance from the sensors
     return annot
+
+
+
+def write_fs_data(p, subjects, structurals, analysis, conditions,
+                  inverse = "meg",
+                  method="dspm", lambda2=0.1111111111111111, pick_ori=None,
+                  n_smooth=25, vertices_to=10242, subjects_dir=None):
+    """Write out group level source data.
+
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


mnefun/_mnefun.py
```diff
@@ -3036,3 +3036,87 @@ def annotate_head_pos(raw, head_pos, rotation_limit=45, translation_limit=0.1,
...
+        # put in big matrix
+        if subj == subjects[0]:
+            data = np.empty((len(stcs), len(subjects), stcs[0].shape[0],
+                             stcs[0].shape[1]))
+        for di, stc in enumerate(stcs):
+            data[di, si, :, :] = stc.data
+            times = stc.times
+    print('Writing data...')
+    np.savez_compressed(fname_data, data=data, times=times, naves=naves)
+    h5f = h5py.File(fname_data, 'w')
```
larsoner: I wouldn't write it this way, but rather using `stc.save(...)` as it's more standard

ktavabi: We could save group averaged data as an `stc` but this writes out individual source data as an ndarray that can later be used for t-testing.

larsoner: I'm saying that individual files should be written for each subject. When t-testing is done you can read data looping over subjects to construct this `ndarray`. It's less convenient but sticks with standard MNE formats. You can always add something like this outside of the `mnefun` scripts. That's how I usually do stats.

ktavabi: Oh... that's how I do it already too. Lets just close.

PR #215, opened 2018-03-13
ENH file checks

ktavabi:
@larsoner please review. Fixes IO error during `gen_report` in case `proj` file(s) are missing.


larsoner:
Thanks @ktavabi

PR #220, opened 2018-04-20
ENH: Add projector averaging

larsoner:
This adds an option to average epochs when computing EOG and ECG projectors. In some testing this seems to work a little bit better.

@ktavabi at some point you might want to try this on your data. In the meantime the default remains `False`, which is the same as it was before.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -425,6 +425,7 @@ class Params(Frozen):
         self.coil_dist_limit = 0.005
         self.coil_t_window = 0.2  # default is same as MF
         self.coil_t_step_min = 0.01
+        self.proj_ave = False
         self.freeze()
 
     @property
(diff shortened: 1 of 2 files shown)


mnefun/_mnefun.py
```diff
@@ -2252,7 +2254,7 @@ def do_preprocessing_combined(p, subjects, run_indices):
                                  tmin=eog_t_lims[0], tmax=eog_t_lims[1],
                                  l_freq=None, h_freq=None, no_proj=True,
                                  ch_name=p.eog_channel,
-                                 reject=p.ssp_eog_reject)
+                                 reject=p.ssp_eog_reject, average=p.proj_ave)
```
ktavabi: So `compute_proj_exg` now computes the projections based on the averaged artifact epochs? What was it doing before?

larsoner: Computing them based on the covariance of the epochs (which is slightly different). In theory if the events are time-locked, which ECG/EOG events appear to be for us, then I think computing it based on the Evoked should work better

ktavabi: Okay. Yeah, I am pretty sure that's how it's benn done in BESA for ever, I just assumed it was done based on averaging. Good to know, I'll try it.

PR #222, opened 2018-04-30
Remove anatomy scripts

ktavabi:
@larsoner we should remove these.


ktavabi:
@larsoner you mind if I merge this?

PR #225, opened 2018-06-06
ENH: Compute noise covar rank during inverse

### LINKED ISSUE #224: On missing epochs key
@larsoner I am hitting a corner case where infrequent event key is missing during epoching and causing a `KeyError`. What is your thought on providing a `params.switch` to change the `on_missing` behavior of `Epochs`?

ktavabi:
@larsoner also fixes #224 by adding docstring.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -208,6 +208,9 @@ class Params(Frozen):
         Low-pass transition band.
     hp_trans : float
         High-pass transition band.
+
+    Attributes
+    ----------
     movecomp : str | None
         Movement compensation to use. Can be 'inter' or None.
     sss_type : str
@@ -249,9 +252,6 @@ class Params(Frozen):
     src_pos : float
         Default is 7 mm. Defines source grid spacing for volumetric source
         space.
-
-    Attributes
-    ----------
     on_missing : string
         Can set to ‘error’ | ‘warning’ | ‘ignore’. Default is 'error'. Determine
         what to do if one or several event ids are not found in the recording
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -249,6 +249,9 @@ class Params(Frozen):
     src_pos : float
         Default is 7 mm. Defines source grid spacing for volumetric source
         space.
+
+    Attributes
+    ----------
     on_missing : string
         Can set to ‘error’ | ‘warning’ | ‘ignore’. Default is 'error'. Determine
         what to do if one or several event ids are not found in the recording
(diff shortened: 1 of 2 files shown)


mnefun/_mnefun.py
```diff
@@ -249,6 +249,13 @@ class Params(Frozen):
     src_pos : float
         Default is 7 mm. Defines source grid spacing for volumetric source
         space.
+    on_missing : string
+        Can set to ‘error’ | ‘warning’ | ‘ignore’. Default is 'error'. Determine
+        what to do if one or several event ids are not found in the recording
+        during epoching. See mne.Epochs docstring for further details.
+    compute_rank : bool
+        Default is False. Set to True to compute rank of the noise covariance
+        matrix during inverse kernel computation.
```
larsoner: The only things that should go here are the arguments of `__init__`. Now would be a good time to add an `Attributes` section to this docstring that includes the things that are available but not in the constructor, starting with these two (can add others later). See:

https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

ktavabi: LMK if the change is what you are referring to. If so, should I just move all the arguments not part of `__init__` into the attributes section?

larsoner: > should I just move all the arguments not part of __init__ into the attributes section?

Yes if you have time please do

mnefun/_mnefun.py
```diff
@@ -250,6 +250,16 @@ class Params(Frozen):
...
+    Attributes
+    ----------
+    on_missing : string
+        Can set to ‘error’ | ‘warning’ | ‘ignore’. Default is 'error'. Determine
+        what to do if one or several event ids are not found in the recording
+        during epoching. See mne.Epochs docstring for further details.
+    compute_rank : bool
+        Default is False. Set to True to compute rank of the noise covariance
+        matrix during inverse kernel computation.
+
```
ktavabi: Is it necessary to redefine parameters as `self.param = arg` after it's been defined in `__init__`? e.g., `self.ssp_eog_reject = ssp_eog_reject`

larsoner: It is necessary to have this assignment in the code, yes, but it only needs to be documented in the docstring in the Parameters section here.

ktavabi: Why are the assignments necessary?

larsoner: If the line `self.ssp_eog_reject = ssp_eog_reject` were omitted, there would be no `params.ssp_eog_reject`. The instance only gets attributes that you assign to it.
```
>>> class Dummy(object)
...    def __init__(self, param)
...        pass
...
>>> d = Dummy('foo')
>>> d.param
  File "<stdin>", line 1, in <module>
AttributeError: 'Dummy' object has no attribute 'param'
```

larsoner:
@mdclarke you might want to start using `compute_rank=True`, it should improve the quality of dSPM for subjects with a lot of head movements

larsoner:
Thanks @ktavabi

PR #228, opened 2018-06-26
Mdclarke reports

mdclarke:


larsoner:
Can you add it to `examples/funloc/analysis_funloc.py`?

ktavabi:
@mdclarke I hit import error on this branch for `plot_snr_estimate` you have to merge #227 with this branch for this to work without error. Also, see if you can squash the verbosity for `plot_snr_estimate` 

@larsoner is there a reason for setting `ylim` to the maxima of snr estimate in the plot? 

screen shot 2018-06-27 at 9 24 34 am

I think the same thing is happening in the whitened evoked plots in mne.

larsoner:
> is there a reason for setting ylim to the maxima of snr estimate in the plot? 

This is probably a bug

> I think the same thing is happening in the whitened evoked plots in mne.

IIRC there it is always set to a max of 10 because you want to see if the baseline stays within the dashed lines at 2 (which it should most of the time)

mdclarke:
@ktavabi ``plot_snr_estimate`` doesn't seem to have a verbose option

mdclarke:
@larsoner ok sorry!

PR #231, opened 2018-09-27
FIX: Fix concat

larsoner:
Modernizes the cov picking and adds some error checking.

@mdclarke can you try it?


mdclarke:
@larsoner yes it is!

PR #233, opened 2018-10-22
Fix Unbound local ref

ktavabi:


mnefun/_mnefun.py
```diff
@@ -1223,7 +1223,7 @@ def _load_trans_to(p, subj, run_indices, raw=None):
                                subj + '_twa_pos.fif')
             if not op.isfile(trans_to):
                 calc_twa_hp(p, subj, trans_to, run_indices)
-        trans_to = mne.read_trans(trans_to)
+        trans_to = mne.read_trans(p.trans_to)
```
larsoner: this isn't the right fix, it should be:
```
else:
    trans_to = p.trans_to
```
Otherwise the conditionals that precede this line are useless

ktavabi:
Look good to you now? @larsoner

larsoner:
The commit history seems to contain things it should not

larsoner:
(the commit itself, rather)

ktavabi:
Jesus, just when you think you know what you're doing :disappointed:

mnefun/_mnefun.py
```diff
@@ -1223,7 +1223,8 @@ def _load_trans_to(p, subj, run_indices, raw=None):
                                subj + '_twa_pos.fif')
             if not op.isfile(trans_to):
                 calc_twa_hp(p, subj, trans_to, run_indices)
-        trans_to = mne.read_trans(trans_to)
+        else:
+            trans_to = mne.read_trans(p.trans_to)
```
larsoner: not quite:
```
else:
    trans_to = p.trans_to
trans_to = mne.read_trans(trans_to)
```

ktavabi: That seems redundant...why redifine `trans_to` if it's in the conditional block?

larsoner: Think about the flow and what happens in all three cases and it should make sense.

Basically the job of the conditional in `master` is to remap `p.trans_to`, which can be a mode *or* filename, to a filename in the variable `trans_to`. After the conditional the `trans` is actually read. The bug is that `trans_to` is not set if `p.trans_to` is a filename. Hence why you need to add a third branch.

In your revision of the code, the `trans_to` is only an instance of `Transformation` in the third branch of the conditional. In the other two, it is a filename. This might work in some situations (as MNE can tolerate filenames well usually) but it's less clean than ensuring the output is already a Transformation.

larsoner:
Thanks @ktavabi

PR #235, opened 2018-11-14
FIX: Fixes for cov and hotelling

larsoner:
Has fixes for latest rank support in MNE, plus a tweak for `hotelling`.

@ktavabi if you have been using `cov_method='shrunk'` it should work better now.

### DIFF
--- a/mnefun/stats.py
+++ b/mnefun/stats.py
@@ -132,7 +132,7 @@ def hotelling_t2_baseline(stc, n_ave, baseline=(None, 0), check_baseline=True):
     return stc
 
 
-def hotelling_t2(epochs, inv_op, baseline=(None, 0)):
+def hotelling_t2(epochs, inv_op, src, baseline=(None, 0), update_interval=10):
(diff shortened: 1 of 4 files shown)


mnefun/stats.py
```diff
@@ -132,7 +132,7 @@ def hotelling_t2_baseline(stc, n_ave, baseline=(None, 0), check_baseline=True):
     return stc
 
 
-def hotelling_t2(epochs, inv_op, baseline=(None, 0)):
+def hotelling_t2(epochs, inv_op, src, baseline=(None, 0), update_interval=10):
     """Compute p values from a VectorSourceEstimate."""
```
ktavabi: What is an example use case for this? Is it only suitable for Vector STC?

ktavabi:
Thanks, @larsoner this LGTM. I am afraid I am not really familiar with all the developments around covariance estimations in MNE so I don't have much to say here. But in general, I think it would be useful to add covariance plots to the report.

PR #243, opened 2019-04-10
ENH: Add HCPMMP mapping

larsoner:
I often point people to this as a ref for the Glasser parcellation

https://images.nature.com/full/nature-assets/nature/journal/v536/n7615/extref/nature18933-s3.pdf

This gives the int->name mapping for this.

Also includes a simplification of the reordering code.


larsoner:
@ktavabi pushed a commit for functional label extraction. Want to see if it would simplify some of your `plot_extract_func_label` code?

larsoner:
This seemed to work well for Melanie's data, so I'll merge. @ktavabi if you try it later and it does not work or do enough for you we can follow up with another PR.

PR #244, opened 2019-05-06
FIX: Rejection criteria bug in raw SSP routine

ktavabi:
Fixes minor bug in passing rejection criteria during SSP computation.


larsoner:
Thanks @ktavabi

PR #249, opened 2019-05-23
ENH: Use cross validated rejection criteria for bad channel selection

ktavabi:
Enable user to use CV rejection criteria in autoreject module during automatic bad channel selection routine:
- Use pre-computed rejection criteria if the container `h5` file corresponding to epochs file exists OR
- Use Autoreject to compute criteria on the fly
-  If neither option is available fall back to the user-defined rejection criteria. 

@larsoner LMKWYT

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -2422,7 +2422,7 @@ def do_preprocessing_combined(p, subjects, run_indices):
             assert p.auto_bad_flat is None or isinstance(p.auto_bad_flat, dict)
             assert p.auto_bad_reject is None or isinstance(p.auto_bad_reject,
                                                            dict) or \
-                   p.auto_bad_reject is 'auto'
+                   p.auto_bad_reject == 'auto'
             if p.auto_bad_reject is 'auto':
                 print('    Auto bad channel selection active. '
                       'Will try using Autoreject module to '
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -2436,18 +2436,16 @@ def do_preprocessing_combined(p, subjects, run_indices):
                                       '     channel selection either define '
                                       '     rejection criteria or install '
                                       '     Autoreject module.\n')
-                else:
-                    print('    Computing thresholds.\n', end='')
-                    temp_epochs = Epochs(
-                            raw, events, event_id=None, tmin=rtmin, tmax=rtmax,
-                            baseline=_get_baseline(p), proj=True, reject=None,
-                            flat=None,
-                            preload=True, decim=1)
-                    kwargs = dict()
-                    if 'verbose' in get_args(get_rejection_threshold):
-                        kwargs['verbose'] = False
-                    reject = get_rejection_threshold(temp_epochs, **kwargs)
-                    reject = {kk: vv for kk, vv in reject.items()}
+                print('    Computing thresholds.\n', end='')
+                temp_epochs = Epochs(
+                    raw, events, event_id=None, tmin=rtmin, tmax=rtmax,
+                    baseline=_get_baseline(p), proj=True, reject=None,
+                    flat=None, preload=True, decim=1)
+                kwargs = dict()
+                if 'verbose' in get_args(get_rejection_threshold):
+                    kwargs['verbose'] = False
+                reject = get_rejection_threshold(temp_epochs, **kwargs)
+                reject = {kk: vv for kk, vv in reject.items()}
             elif p.auto_bad_reject is None and p.auto_bad_flat is None:
                 raise RuntimeError('Auto bad channel detection active. Noisy '
                                    'and flat channel detection '
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -2405,6 +2405,10 @@ def do_preprocessing_combined(p, subjects, run_indices):
                              pick=True, skip_by_annotation='edge',
                              **fir_kwargs)
             events = fixed_len_events(p, raw)
+            rtmin = p.reject_tmin \
+                if p.reject_tmin is not None else p.tmin
+            rtmax = p.reject_tmax \
+                if p.reject_tmax is not None else p.tmax
             # do not mark eog channels bad
             meg, eeg = 'meg' in raw, 'eeg' in raw
             picks = pick_types(raw.info, meg=meg, eeg=eeg, eog=False,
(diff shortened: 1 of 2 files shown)


larsoner:
I'm not sure about that flow/logic, if a user supplies `auto_bad_reject = dict(...)` then it should be used and `autoreject` (and files) should never be used. Maybe `auto_bad_reject = 'auto'` should trigger this automatic mode?

There is code to look for some file, but not to write one out. Is this intentional? If `autoreject` does not take too long to run (e.g., less than one minute) then I probably wouldn't bother with files. If you do want to have files, then `mnefun` it should probably write them out. If you're thinking that the user might write out their own files, this is probably better supported with a `defaultdict` specified in the analysis script.

ktavabi:
> if a user supplies auto_bad_reject = dict(...)

This only seems to work reasonably for adult recordings. 

> Maybe auto_bad_reject = 'auto' should trigger this automatic mode

That’s a possibility 

> There is code to look for some file, but not to write one out. Is this intentional?

Yes. I noticed that you started writing out the rejection  criteria as age 5 files, and use them during covariance routine. So I figured using them to save time might make sense. But generally auto reject routine does not take too long and could be used on the fly  although that would mean repeating the operation

larsoner:
>> if a user supplies auto_bad_reject = dict(...)

> This only seems to work reasonably for adult recordings.

That's fine, we can provide a way to use `autoreject` but from an API perspective if someone does `auto_bad_reject = dict` it should be used (regardless of recording type). The 'auto' option is an explicit way to make it use autoreject to set the rejection thresholds for automatic bad channel detection.

> I noticed that you started writing out the rejection criteria as age 5 files, and use them during covariance routine.

I see. But I think those are for *after* SSP and bad channel removal, right? So they probably aren't appropriate here.

ktavabi:
@larsoner so far this appears to work for infant MEG. _funloc_ MEEG data failed with >10 EEG channels marked bad. I am wondering if it makes any sense to scale computed rejection criteria e.g., reject if exceeding twice computed criteria. LMKWYT.

larsoner:
`autoreject` has some strictness criteria, right? If so, can that be adjusted?

ktavabi:
Not explicitly. But there is an `RMSE` we could theoretically use (http://autoreject.github.io/auto_examples/plot_global_reject.html#sphx-glr-auto-examples-plot-global-reject-py)

larsoner:
If it works for infant data but not for adults (or maybe just not adults with EEG?) we can just make note of this

ktavabi:
> If it works for infant data but not for adults (or maybe just not adults with EEG?) we can just make note of this

Make a note of `RMSE` feature of autoreject? Or scaling the computed thresh? You are ok to leave in the scaling fixed at `2x` for now? 
I'll dink around with using RMSE values from Autoreject in a future PR.

larsoner:
> Make a note of RMSE feature of autoreject? Or scaling the computed thresh?

Make a note along the lines of our discussion above that "if a user supplies auto_bad_reject = dict(...) ... This only seems to work reasonably for adult recordings.".

> You are ok to leave in the scaling fixed at 2x for now?

Yeah whatever works well for the data in your tests is fine by me. But let's leave this stuff disabled by default for backward compat (and for adult recordings).

ktavabi:
I'll leave further mods contingent on results from #251 which should affect rejection criteria.

larsoner:
Feel free to try now

ktavabi:
I am running about 2 hundred datasets through with MNEFUN now. About 6% come back with >15 bad channels with MF autobad, a subsequent 3% with too many bad channels after autoreject thresholding without the scaling. I am not really fond of this arbitrary scaling, and these numbers I can live with. 

I'll push changes Mon. to 

- [ ] tidy up messaging
- [ ] finish up with documentation

ktavabi:
@larsoner does the commit history for this look "reasonable"? I can't tell what is happening on my end. Closes #249

mnefun/_mnefun.py
```diff
@@ -183,6 +183,13 @@ class Params(Frozen):
     auto_bad : float | None
         If not None, bad channels will be automatically excluded if
         they disqualify a proportion of events exceeding ``autobad``.
+    auto_bad_reject : str | dict
+        Default is None. Must be defined if using Autoreject module to
+        compute noisy sensor rejection ctriteria. Set to 'auto' to compute
```
larsoner:
```suggestion
        compute noisy sensor rejection criteria. Set to 'auto' to compute
```

mnefun/_mnefun.py
```diff
@@ -2405,24 +2411,55 @@ def do_preprocessing_combined(p, subjects, run_indices):
...
-            if p.auto_bad_reject is None and p.auto_bad_flat is None:
+                                                           dict) or \
+                   p.auto_bad_reject is 'auto'
+            if p.auto_bad_reject is 'auto':
+                print('    Auto bad channel selection active. '
+                      'Will try using Autoreject module to '
+                      'compute rejection criterion.\n' , end='')
+                try:
+                    from autoreject import get_rejection_threshold
+                    print('    Computing autoreject thresholds.\n', end='')
```
larsoner: Only put in `try` the minimal set of things expected to cause the error. So here do:
```
try:
    from autoreject import ...
except ImportError:
    print(...)
else:
    print(...)
    temp_epochs = ...
```
but actually if someone requests it but does not have it I would raise an error:
```
try:
   from autoreject import ...
except ImportError:
   raise ImportError(...)
print(...)
temp_epochs = ...
```

mnefun/_mnefun.py
```diff
@@ -2405,24 +2411,55 @@ def do_preprocessing_combined(p, subjects, run_indices):
...
             meg, eeg = 'meg' in raw, 'eeg' in raw
             picks = pick_types(raw.info, meg=meg, eeg=eeg, eog=False,
                                exclude=[])
             assert p.auto_bad_flat is None or isinstance(p.auto_bad_flat, dict)
             assert p.auto_bad_reject is None or isinstance(p.auto_bad_reject,
-                                                           dict)
-            if p.auto_bad_reject is None and p.auto_bad_flat is None:
+                                                           dict) or \
+                   p.auto_bad_reject is 'auto'
+            if p.auto_bad_reject is 'auto':
```
larsoner:
```suggestion
            if p.auto_bad_reject == 'auto':
```
do not use `is` with `str` comparisons, it can break on different encodings

mnefun/_mnefun.py
```diff
@@ -2405,24 +2411,55 @@ def do_preprocessing_combined(p, subjects, run_indices):
...
             # do not mark eog channels bad
             meg, eeg = 'meg' in raw, 'eeg' in raw
             picks = pick_types(raw.info, meg=meg, eeg=eeg, eog=False,
                                exclude=[])
             assert p.auto_bad_flat is None or isinstance(p.auto_bad_flat, dict)
             assert p.auto_bad_reject is None or isinstance(p.auto_bad_reject,
-                                                           dict)
-            if p.auto_bad_reject is None and p.auto_bad_flat is None:
+                                                           dict) or \
+                   p.auto_bad_reject is 'auto'
```
larsoner:
```suggestion
                   p.auto_bad_reject == 'auto'
```

mnefun/_mnefun.py
```diff
@@ -183,6 +183,13 @@ class Params(Frozen):
     auto_bad : float | None
         If not None, bad channels will be automatically excluded if
         they disqualify a proportion of events exceeding ``autobad``.
+    auto_bad_reject : str | dict
```
larsoner:
```suggestion
    auto_bad_reject : str | dict | None
```

mnefun/_mnefun.py
larsoner: This was marked resolved but was not fixed

mnefun/_mnefun.py
```diff
@@ -2405,24 +2411,57 @@ def do_preprocessing_combined(p, subjects, run_indices):
...
             assert p.auto_bad_flat is None or isinstance(p.auto_bad_flat, dict)
             assert p.auto_bad_reject is None or isinstance(p.auto_bad_reject,
-                                                           dict)
-            if p.auto_bad_reject is None and p.auto_bad_flat is None:
+                                                           dict) or \
+                   p.auto_bad_reject == 'auto'
+            if p.auto_bad_reject is 'auto':
+                print('    Auto bad channel selection active. '
+                      'Will try using Autoreject module to '
+                      'compute rejection criterion.\n' , end='')
```
larsoner:
```suggestion
                      'compute rejection criterion.')
```

mnefun/_mnefun.py
```diff
@@ -2405,24 +2411,57 @@ def do_preprocessing_combined(p, subjects, run_indices):
...
+                    from autoreject import get_rejection_threshold
+                except ImportError:
+                    raise ImportError('     Autoreject module not installed.\n'
+                                      '     Noisy channel detection parameter '
+                                      '     not defined. To use autobad '
+                                      '     channel selection either define '
+                                      '     rejection criteria or install '
+                                      '     Autoreject module.\n')
+                else:
+                    print('    Computing thresholds.\n', end='')
```
larsoner: No need to nest these lines in an `else` clause anymore (they will only execute if there was no `ImportError` either way)

mnefun/_mnefun.py
larsoner: This is still wrong

larsoner:
Looks reasonable to me. I have not tested but I assume you will report bugs if it does not work :)

Issue #250, opened 2019-05-30
ENH: MF autobad


ktavabi:
Add option to run Maxfilter for noisy channel detection prior to mne-python SSS routines.

PR #251, opened 2019-05-30
WIP Add Maxfilter autobad channel detection

ktavabi:
WIP to address #250 @larsoner please review. 
- [x] backup prebad file.
- [x] write MF noisy channel list to file.
- [x] modify  `run_sss_command` to run MF.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -1016,8 +1016,9 @@ def fetch_sss_files(p, subjects, run_indices):
 
 
 def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
-                    fname_pos=None, stdout=None, stderr=None, prefix='',
-                    work_dir='~/'):
+                    fname_pos=None, fname_bads=None, stdout=None, stderr=None,
+                    prefix='',
+                    work_dir='~/', **kwargs):
     """Run Maxfilter remotely and fetch resulting file
 
     Parameters
@@ -1033,6 +1034,8 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
         The SSH/scp host to run the command on.
     fname_pos : str | None
         The ``-hp fname_pos`` to use with MaxFilter.
+    fname_bads : str | None
+        The output file containing channels detected by MaxFilter as bad.
     stdout : file-like | None
         Where to send stdout.
     stderr : file-like | None
@@ -1041,12 +1044,15 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
         The text to prefix to messages.
     work_dir : str
         Where to store the temporary files.
+    **kwargs: dict
+        Additional command line arguments to Maxfilter as key, value pairs.
     """
     # let's make sure we can actually write where we want
     if not op.isfile(fname_in):
         raise IOError('input file not found: %s' % fname_in)
-    if not op.isdir(op.dirname(op.abspath(fname_out))):
-        raise IOError('output directory for output file does not exist')
+    if fname_out is not None:
+        if not op.isdir(op.dirname(op.abspath(fname_out))):
+            raise IOError('output directory for output file does not exist')
     if any(x in options for x in ('-f ', '-o ', '-hp ')):
         raise ValueError('options cannot contain -o, -f, or -hp, these are '
                          'set automatically')
@@ -1057,9 +1063,21 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
     remote_pos = op.join(work_dir, 'temp_%s_raw_sss.pos' % t0)
     print('%sOn %s: copying' % (prefix, host), end='')
     fname_in = op.realpath(fname_in)  # in case it's a symlink
-    cmd = ['scp', '-P' + port, fname_in, host + ':' + remote_in]
+    cmd = ['scp', '-P ' + port, fname_in, host + ':' + remote_in]
     run_subprocess(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-
+    
+    if fname_bads is not None:
+        if kwargs is not None:
+            if 'badlimit' in kwargs.keys():
+                options += ' -badlimit %d' % kwargs['badlimit']
+        options += ' -autobad on -skip 0 30'
+        cmd = ['ssh', '-p', port, host,
+               'maxfilter -f ' + remote_in + ' -o ' + remote_out + ' ' +
+               options]
+        with open(fname_bads, "w+") as f:
+            run_subprocess(cmd, stdout=f, stderr=subprocess.PIPE)
+        options = '-v -force'
+    
     if fname_pos is not None:
         options += ' -hp ' + remote_pos
 
@@ -1068,8 +1086,8 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
            'maxfilter -f ' + remote_in + ' -o ' + remote_out + ' ' + options]
     try:
         run_subprocess(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-
-        print(', copying to %s' % (op.basename(fname_out),), end='')
+        if fname_out is not None:
+            print(', copying to %s' % (op.basename(fname_out),), end='')
         if fname_pos is not None:
             try:
                 cmd = ['scp', '-P' + port, host + ':' + remote_pos, fname_pos]
@@ -1077,8 +1095,6 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
                                stderr=subprocess.PIPE)
             except Exception:
                 pass
-        cmd = ['scp', '-P' + port, host + ':' + remote_out, fname_out]
-        run_subprocess(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     finally:
         print(', cleaning', end='')
         files = [remote_in, remote_out]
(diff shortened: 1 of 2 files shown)


mnefun/_mnefun.py
```diff
@@ -262,6 +262,12 @@ class Params(Frozen):
         Default is ('mag', 'grad', 'eeg'). Can set to ('mag', 'grad', 'eeg',
         'eog) to use EOG channel rejection criterion from autoreject module to
         reject trials on basis of EOG.
+    mf_autobad : bool
+        Default False. If True use Maxfilter to automatically mark bad
+        channels prior to SSS denoising.
+    mf_badlimit : int
+        Default is 7. Threshold limit for Maxfilter noisy channel
+        detection.
```
larsoner: What is the default value in `maxfilter`? We should use whatever their default is, probably.

mnefun/_mnefun.py
```diff
@@ -1246,6 +1254,13 @@ def run_sss_locally(p, subjects, run_indices):
             raw = read_raw_fif(r, preload=True, allow_maxshield='yes')
             raw.fix_mag_coil_types()
             _load_meg_bads(raw, prebad_file, disp=ii == 0, prefix=' ' * 6)
+            # If so run Maxfilter for automatic bad channel selection
+            if p.mf_autobad:
+                print('     Maxfilter auto bad channel selection requested. '
+                      'Running on kasga...', end='')
+                add_bads = do_mf_autobad(p, r, prebad_file)
```
larsoner: I think this should write to a file, and only run `maxfilter` if the file does not exist. It makes it so that you only ever need to run it once, even if you run `sss` multiple times.

mnefun/_mnefun.py
```diff
@@ -3650,3 +3686,54 @@ def extract_roi(stc, src, label=None, thresh=0.5):
...
     func_label = func_label.fill(src)
     return func_label, max_vert, max_vidx, max_tidx
+
+
+def do_mf_autobad(p, r, prebad_file):
+    t0 = time.time()
+    tmp_name = op.join(op.dirname(r), 'temp_%s_raw.fif' % t0)
+    tmp_raw = filter_chpi(read_raw_fif(r, allow_maxshield='yes',
+                                       preload=True),
+                          include_line=True)
```
larsoner: Why this step? Is it necessary? I think back in the day when we did this before with shell scripts we did not do cHPI filtering, or if we did, we let MaxFilter do it while processing the data.

mnefun/_mnefun.py
```diff
@@ -3650,3 +3686,54 @@ def extract_roi(stc, src, label=None, thresh=0.5):
...
+    remote_out = op.join(p.sws_ssh, p.sws_dir,
+                         op.basename(tmp_name))
+    mf_log = op.join(p.sws_ssh, p.sws_dir,
+                     '%s_stdout.txt' % op.basename(remote_out[:-8]))
+    cmd = ['ssh', '-p', str(p.sws_port), p.sws_ssh,
+           '/neuro/bin/util/maxfilter -f ' + remote_in + ' -o ' +
+           remote_out + ' -autobad on -skip 0 30 -format short '
+                        '-badlimit %d -v' % p.mf_badlimit]
+    output = run_subprocess(cmd, stdout=subprocess.PIPE,
+                            stderr=subprocess.DEVNULL)[0]
```
larsoner: Can you make use of `run_sss_command` to avoid this boilerplate ssh-run-copy-delete stuff? You might have to modify `run_sss_command` a bit to get it to do what you want

mnefun/_mnefun.py
ktavabi: Okay

mnefun/_mnefun.py
ktavabi: 7 is Default. I used five the test.

mnefun/_mnefun.py
ktavabi: I wasn’t aware if Maxfilter does cHPI filtering. If so then we can leave the step out.

larsoner: I think it does by default. Most of the time people I've seen people use `autobad` they don't manually remove cHPI themselves before doing it

mnefun/_mnefun.py
ktavabi: Initially tried getting `run_sss_comand` to do this but wasn't able to. I could give it another go but wasn't sure if you wanted the function modified and it was relatively less pain with `subprocess`.

larsoner: I'd rather get run_sss_command to have the right options to get what you need. That way the pushing, running, and deleting can be in one place

mnefun/_mnefun.py
```diff
@@ -1049,9 +1063,21 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
...
+        if kwargs is not None:
+            if 'badlimit' in kwargs.keys():
+                options += ' -badlimit %d' % kwargs['badlimit']
+        options += ' -autobad on -skip 0 30'
+        cmd = ['ssh', '-p', port, host,
+               'maxfilter -f ' + remote_in + ' -o ' + remote_out + ' ' +
+               options]
+        with open(fname_bads, "w+") as f:
+            run_subprocess(cmd, stdout=f, stderr=subprocess.PIPE)
+        options = '-v -force'
```
larsoner: I don't think this is what we want to do. I think `run_sss_command` should always:

1. `rsync` to push to remote
2. run one `maxfilter` call remotely
3. retrieve any files or outputs
4. clean up the remote (remove the file)

So if in their analysis script a user has both `mf_autobad=True` and `movecomp=True`, it means you will go through this cycle twice (independently).

If you're worried about efficiency, we can make step (4) optional with a `remove_transferred=True`. If the user has `p.movecomp` then in the "mf_autobad" step we can call `remove_transferred=False`. And when movecomp is done it won't need to re-transfer.

mnefun/_mnefun.py
```diff
@@ -2750,6 +2802,45 @@ def _prebad(p, subj):
...
+            output = [line for line in f]
+        # Parse output for bad channels
+        if any('Static bad channels' in ss for ss in output):
+            chs = [ll for ll in output if 'Static bad channels' in ll][
+                0].split(':')[1].strip()
+            with open(bad_file, "w+") as f:
+                f.write(chs)
+        print('Backing-up %s.orig: ' % prebad_file,
+              '\nWriting modified %s: ' % prebad_file, end='')
+        copy2(prebad_file, prebad_file + '.orig')
```
larsoner: I would not overwrite anything, but instead when we want to use the prebad/maxbad files (always together, probably only in the Maxfiltering step):

1. Read the channels from the manually created `_prebad.txt` file (if it exists)
2. Read the channels from the MF-created `_maxbad.txt` file
3. Concat these in memory like `all_bads = sorted(set(prebads) + set(maxbads))`

Then use `raw.info['bads'] = all_bads` or so

ktavabi:
This is ready to merge from my end.

larsoner:
I made comments yesterday, I think perhaps you missed them?

mnefun/_mnefun.py
```diff
@@ -1049,9 +1063,21 @@ def run_sss_command(fname_in, options, fname_out, host='kasga', port=22,
     remote_pos = op.join(work_dir, 'temp_%s_raw_sss.pos' % t0)
     print('%sOn %s: copying' % (prefix, host), end='')
     fname_in = op.realpath(fname_in)  # in case it's a symlink
-    cmd = ['scp', '-P' + port, fname_in, host + ':' + remote_in]
+    cmd = ['scp', '-P ' + port, fname_in, host + ':' + remote_in]
     run_subprocess(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-
+
```
ktavabi: >`rsync` to push to remote
We already using `scp` to push files to remote.

mnefun/_mnefun.py
ktavabi: > 1. rsync to push to remote
> 2. run one maxfilter call remotely
> 3. retrieve any files or outputs

At the moment I am doing just one `maxfilter` call remotely and piping the `stdout` to a local file i.e., not writing to remote file and rsyncing, which is rather nice. The alternative requires piping the MF shell output to a remote file before pulling, and I couldn't figure out how to do do that with `run_subprocess`. 

> So if in their analysis script a user has both mf_autobad=True and movecomp=True, it means you will go through this cycle twice (independently).

Not exactly sure WYM here. At the moment the HP estimation step for movecomp is being handled by `run_sss_positions` so they are independent.

> 4.clean up the remote (remove the file)
At the moment the clean up is working just fine without any changes to `run_sss_command`.  Or do you mean something else?

> If the user has p.movecomp then in the "mf_autobad" step we can call remove_transferred=False. And when movecomp is being executed it won't need to re-transfer.

If I am understanding this correctly, we may have to modify 2-3 functions to track files with tmp names as per `time.time` to save a few minutes repeating file transfers.

mnefun/_mnefun.py
ktavabi: Actually, the default value 7 seems to work reasonably well for infant data, maybe just get rid of it?

larsoner: Sure

mnefun/_mnefun.py
larsoner: It looks like this has not been addressed

ktavabi: I’ll get to it today

mnefun/_mnefun.py
larsoner: > At the moment I am doing just one maxfilter call remotely and piping the stdout to a local file i.e., not writing to remote file and rsyncing, which is rather nice. The alternative requires piping the MF shell output to a remote file before pulling, and I couldn't figure out how to do do that with run_subprocess.

Conceptually and code wise it makes things less clean to shoe-horn this extra `autobad` step in what is meant to be a generic `maxfilter` call-and-retrieve function.

By default the output of `run_subprocess` is the `stdout` of the process that was executed, so it should not be so difficult to get the output as a `str`:

https://github.com/mne-tools/mne-python/blob/06c7785aabe72fc7394cc6e6ba1c7d5aec1d2c76/mne/utils/misc.py#L85

I can push some commits to your branch to do this if you prefer?

> If I am understanding this correctly, we may have to modify 2-3 functions to track files with tmp names as per time.time to save a few minutes repeating file transfers.

Okay to keep things simpler, let's not worry about efficiency for now, and just do the `autobad` step in one call and the `headpos` in another.

ktavabi: Sorry dude but I am still not sure WYM WRT to combining the `autobad` & `headpos` routines. Feel free to push to the branch.

larsoner:
@ktavabi sorry I misunderstood a bit what you were doing. I pushed a commit to simplify the handling can you see if the code still makes sense and works for you?

ktavabi:
Hmmm...It appears to be working but it doesn't look like much of anything I was doing :disappointed:

larsoner:
I refactored it in a few ways:

1. Modify `run_sss_command` to return the output rather than redirect stdout in the remote command
2. Move file / output parsing to the `_maxbad` function rather than having it be in `run_sss_command`
3. Made the `maxbad` a per-file thing, rather than global to the subject
4. Didn't overwrite the manual (by eye) `_prebad.txt`
5. Changed `scp` calls to `rsync` in case we eventually want to make it more efficient

So it looks a bit different but should still do more or less what you had before (except point (3) and somewhat (4) which make it operationally a little bit different)

ktavabi:
LGTM! and it's much cleaner

mnefun/_mnefun.py
```diff
@@ -1417,27 +1417,6 @@ def _load_meg_bads(raw, prebad_file, disp=True, prefix='     ',
     raw.info._check_consistency()
 
 
-def _configure_ch_names(ch_names, info):
```
ktavabi: Cool! I was just looking for this one...it's not needed.

PR #259, opened 2019-09-05
ENH: Add drop_log plotting

larsoner:
Adds `drop_log` plotting and fixes a problem with `fit_data` being `None` in report.

### DIFF
--- a/examples/funloc/analysis_fun.py
+++ b/examples/funloc/analysis_fun.py
@@ -155,16 +155,24 @@ params.must_match = [
 ]
 
 params.report_params.update(  # add a couple of nice diagnostic plots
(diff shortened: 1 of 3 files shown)


larsoner:
@ktavabi feel free to review when you get a chance

ktavabi:
LGTM. Do you want to also add the option to the example script?

larsoner:
It's enabled by default so shouldn't need to do anything to get it to show up. I guess it could be added to expose the option, is that what you had in mind?

ktavabi:
Yeah, that is what I was referring to.

larsoner:
> Also, would this be considered DRY?

One is for during-processing and the other is for the report. I think they serve different enough purposes to keep them separate.

PR #264, opened 2019-10-16
FIX: Minor fixes

### LINKED ISSUE #263: want_rank != out_rank
in the `generate_covariances` step, @nordme and I are hitting an `AssertionError` where `want_rank != out_rank`.  The one we just looked at was off by one, but I haven't yet looked at all the failing subjects (there are several) to confirm that it's always just a difference of one (which might suggest something as simple as a tolerance issue?).

I dug into it a little:

- `want_rank` -> `_mnefun.compute_rank` -> `mne.rank.estimate_rank` (with tolerance 1e-6) -> `_estimate_rank_from_s` (`s` comes from `scipy.linalg.svdvals`; rank = `sum(s > tolerance)`)
- `out_rank` -> `mne.cov.compute_whitener` -> `mne.cov.prepare_noise_cov` -> `_smart_eigh` (rank = `sum(eig > 0)`)

@larsoner what is your advice on how to fix / work around this issue?

larsoner:
1. Unifies how `epochs` are created in `save_epochs` and in `gen_covariances`. Critically the `reject_by_annotation` param differed, which seemed to be breaking the rank estimation.
2. Cleans up the decisions about when the annotation code prints.
3. Fixes a documentation bug about `int_order` vs `ext_order` for infant data.
4. Allows setting `params.cov_rank_tol` to something higher than the default (1e-6).

@nordme @drammock can you see if (1) this fixes your error?

Even after (1), though, I found that some conditions had a bit of a whitening problem in the reports, which could be fixed by (4), namely setting `params.cov_rank_tol = 1e-2`. This seems a bit high, though, so alternatively using `params.cov_method='shrunk'` (instead of `'empirical'`) might be a better choice, as it also fixes the problem and is probably more robust than trying to tweak the cutoff value.

Also you probably want to add `params.proj_meg = 'combined'` to your processing scripts because it works a bit better.

Closes #263.

### DIFF
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -688,10 +688,14 @@ def do_processing(p, fetch_raw=False, do_score=False, push_raw=False,
 
     run_indices = p.subject_run_indices
     if run_indices is None:
-        run_indices = [None] * len(p.subjects)
-    assert len(run_indices) == len(p.subjects)
-    run_indices = [np.array(run_indices[si]) if run_indices[si] is not None
-                   else np.arange(len(p.run_namus)for si in sinds]
+        run_indices = [None] * len(subjects)
+    elif isinstance(run_indices, dict):
+        run_indices = [run_indices[subject] for subject in subjects]
+    else:
+        run_indices = [run_indices[si] for si in sinds]
+        assert len(run_indices) == len(p.subjects)
+    run_indices = [np.array(run_idx) if run_idx is not None
+                   else np.arange(len(p.run_names)) for run_idx in run_indices]
     assert all(run_idx.ndim == 1 for run_idx in run_indices)
     assert all(np.in1d(r, np.arange(len(p.run_names))).all()
                for r in run_indices)
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -238,9 +238,9 @@ class Params(Frozen):
         signal space separation method. Must be either 'maxfilter' or 'python'
     int_order : int
         Order of internal component of spherical expansion. Default is 8.
+        Value of 6 recomended for infant data.
     ext_order : int
         Order of external component of spherical expansion. Default is 3.
-        Value of 6 recomended for infant data
     tsss_dur : float | None
         Buffer length (in seconds) fpr Spatiotemporal SSS. Default is 60.
         however based on system specification a shorter buffer may be
@@ -713,6 +713,8 @@ def do_processing(p, fetch_raw=False, do_score=False, push_raw=False,
             elif func == print_proc_status:
                 outs[ii] = func(p, subjects, structurals, p.analyses,
                                 run_indices)
+            elif func == gen_covariances:
+                outs[ii] = func(p, subjects, run_indices, decim)
             else:
                 outs[ii] = func(p, subjects, run_indices)
             print('  (' + timestring(time.time() - t0) + ')')
@@ -1247,7 +1249,7 @@ def run_sss_positions(fname_in, fname_out, host='kasga', opts='-force',
             print('\n%s... found %d invalid position%s from MaxFilter, '
                   'replacing with: %r'
                   % (prefix, first_valid, _pl(first_valid),
-                     this_pos[first_valid].tolist()), end='')
+                     repl[1:].tolist()), end='')
             this_pos[:first_valid, 1:] = repl[1:]
             next_pre = '\n%s... ' % (prefix,)
         data.append(this_pos)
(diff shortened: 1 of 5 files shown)


nordme:
Hi @larsoner, it looks like the _read_events function is having trouble during the epoching step in scenarios where there's only one raw file per subject. It's expecting to concatenate, but there's nothing for it to work on. 

```
  File "/home/nordme/Documents/prek_mnefun.py", line 168, in <module>
    print_status=True # show status
  File "/home/nordme/GitHub/mnefun/mnefun/_mnefun.py", line 712, in do_processing
    p.must_match, decim, run_indices)
  File "/home/nordme/GitHub/mnefun/mnefun/_mnefun.py", line 1814, in save_epochs
Current compensation grade : 0
    events = _read_events(p, subj, run_indices[si], raw)
  File "/home/nordme/GitHub/mnefun/mnefun/_mnefun.py", line 1707, in _read_events
    assert ridx.ndim == 1
AssertionError
```

larsoner:
It's supposed to be a list per subject, so you probably need `params.subject_indices = [[0]]` for a single subject as opposed to `[0]`

nordme:
Alright, tested against Prek subjects, and it looks good!

larsoner:
Okay great, @drammock can you have a quick look?

larsoner:
@nordme feel free to review the code as well

drammock:
Sure, I can look later this afternoon.

mnefun/_mnefun.py
```diff
@@ -2268,23 +2285,24 @@ def gen_covariances(p, subjects, run_indices):
...
                 print('  Using %s/%s events for %s'
                       % (new_count, old_count, op.basename(cov_name)))
-            events = concatenate_events(events, first_samps,
-                                        last_samps)
             use_reject, use_flat = _restrict_reject_flat(reject, flat, raw)
             epochs = Epochs(raw, events, event_id=None, tmin=p.bmin,
                             tmax=p.bmax, baseline=(None, None), proj=False,
-                            reject=use_reject, flat=use_flat, preload=True)
+                            reject=use_reject, flat=use_flat, preload=True,
+                            decim=decim[si], verbose='error',    # ignore decim
```
drammock:
```suggestion
                            decim=decim[si],
                            verbose='error',  # ignore decim-related warnings
```

mnefun/_mnefun.py
```diff
@@ -2293,7 +2311,31 @@ def gen_covariances(p, subjects, run_indices):
...
+                                          epochs2.events[:, 0])
+                    epochs2.pick_types(meg=True, eeg=True, exclude=[])
+                    import matplotlib.pyplot as plt
+                    plt.figure()
+                    for eps in (epochs, epochs2):
+                        eps = eps.get_data().transpose([1, 0, 2])
+                        eps = eps.reshape(len(eps), -1)
+                        plt.plot(np.log10(np.maximum(linalg.svdvals(eps), 1e-50)))
+                    epochs.plot()
+                    epochs2.copy().crop(p.bmin, p.bmax).plot()
```
drammock: will this automatically `plt.show()`?  (are we guaranteed to be in `ion` mode?)

mnefun/misc.py
```diff
@@ -10,33 +10,37 @@ def make_montage(info, kind, check=False):
...
-    montage.pos *= sphere.radius
-    montage.pos += sphere['r0']
-    info = mne.pick_info(info, picks)
+        assert len(to_names) == 60
+        from_names = getattr(_reorder, 'ch_names_' + kind)
+        montage = mne.channels.make_standard_montage(
+            'standard_1020', head_size=sphere.radius)
+    assert len(from_names) == len(to_names)
+    montage_pos = montage._get_ch_pos()
+    montage = mne.channels.make_dig_montage(
```
drammock: this change requires 0.19 right? Is that assumed / asserted / mentioned somewhere? (sorry if the answer is obvious; this is the first time I'm dabbling with mnefun in several years)

mnefun/_mnefun.py
larsoner: This is just debugging code we will hopefully never need. I left it in just to make life easier when we do. But even then the default of `epochs.plot` has `show=True`

mnefun/misc.py
larsoner: The general assumption is that people using `mnefun` are on a (near-)latest `master` version because often we have to fix bugs. So 0.19 should be a safe enough assumption. `make_dig_montage` didn't exist before then so it will just bomb if someone tries this on 0.18.

Additionally also this function isn't really used in `mnefun` analysis scripts by users, it's more of an expert `dev` option for testing new algorithms. So should be okay

PR #266, opened 2019-11-13
WIP, DOC: Document flow

larsoner:
Let's actually build and deploy docs so that people can read them.


larsoner:
Up at https://labsn.github.io/mnefun

PR #268, opened 2019-11-14
ENH: Modernize flow chart

larsoner:


larsoner:
@drammock @nordme @pettetmw @mdclarke can you look and see if this makes sense:

https://32-17719360-gh.circle-artifacts.com/0/html/overview.html

It's currently a bit thin (see Note at the top, it's basically a TODO list), but I think at least it's accurate about what happens nowadays. Follow-up PRs welcome to extend and improve this if people are already aware of where/how stuff happens. 

@drammock note that the edge names are clickable links :)

larsoner:
I'll merge so we can continue iterating but please do still look when you get a chance

PR #270, opened 2019-11-18
ENH: Use YAML

larsoner:
Todo:

- [x] Get all params in `YAML` and in `Params` docstring
- [x] Add consistency tests
- [x] Organize the YAML
- [x] Document all parameters
- [x] Get converter to actually work
- [x] Move text from `funloc` example to YAML file, Attributes, or Overview section of docs
- [x] Make a canonical, same-as-Params default YAML file
- [x] Use canonical file for validation of YAML entries
- [x] Test `funloc` file is a subset of canonical one
- [x] Add `Params.save` YAML writer (facilitate transition)
- [x] Test YAML I/O round-trip


examples/funloc/funloc_params.yml
```diff
@@ -0,0 +1,138 @@
...
+sss_type: 'python'
+int_order: 8
+ext_order : 3
+tsss_dur: 60.
+st_correlation: 0.98
+trans_to: 'median'
+sss_origin: 'auto'
+fir_design: 'firwin'
+autoreject_thresholds: False
+autoreject_types: ('mag', 'grad', 'eeg')
```
drammock: tuples in YAML source won't work, only lists. this will get imported as `"('mag', 'grad', 'eeg')"`

mnefun/tests/test_documented.py
```diff
@@ -0,0 +1,37 @@
...
+    doc = docscrape.ClassDoc(Params)
+    doc = [p.name for p in doc['Attributes']]
+    assert set(doc) == key_set
+
+    # funloc documented the same way
+    fname = op.join(op.dirname(__file__), '..', '..', 'examples', 'funloc',
+                    'funloc_params.yml')
+    assert op.isfile(fname)  # only works on dev-installed mnefun
+    with open(fname, 'r') as fid:
+        yvals = yaml.safe_load(fid)
```
drammock: more forward-compatible (I think) is `yaml.load(fid, Loader=yaml.SafeLoader)`

examples/funloc/analysis_fun.py
```diff
@@ -32,12 +32,10 @@
 from score import score
 import numpy as np
 
-try:
-    # Use niprov as handler for events if it's installed
-    from niprov.mnefunsupport import handler
-except ImportError:
-    handler = None
+params = mnefun.read_params('funloc_params.yml')
```
drammock: where is `read_params` defined?

larsoner: It's not yet, consider this TDD :)

larsoner:
Okay @drammock I've made enough progress on this that it could use another set of eyes. Basically see if you think the yaml file organization is reasonable. Once it is, I'll check off the rest of the boxes here and then we can hopefully merge.

One question is where to document options. In the current state of things, `Params` has all options listed with some minimal explanations. We have slightly more explanations available in `master` in `examples/funloc/analysis_fun.py`. I could either move those to the `Params` docstring, or I could put them in the YAML (sort of how I did for the `report` options). Where do you think these should live?

larsoner:
... or maybe the best place to put real descriptions is in the `Overview` section of the docs with the flow chart.

examples/funloc/analysis_fun.py
```diff
@@ -26,173 +24,32 @@
...
     * http://lester.ilabs.uw.edu/files/AKCLEE_110_slim.tar.gz
 
-"""
+Setting up a config file
+------------------------
+Parameters for remotely connecting to SSS workstation ('sws') can be set
+by adding a file ~/.mnefun/mnefun.json with contents like:
+
+    $ mkdir ~/.mnefun
+    $ echo '{"sws_ssh":"kasga","sws_dir":"/data06/larsoner/sss_work","sws_port":22}' > ~/.mnefun/mnefun.json
```
drammock: spaces after colons & commas?

drammock: also wondering why this is kept in a JSON file separate from the YAML params...  is it because the YAML is meant to be published with the code, whereas this should not?

examples/funloc/funloc_params.yml
```diff
@@ -0,0 +1,189 @@
...
+    n_jobs_resample: 1
+  bads:
+    mf_autobad: True
+    mf_badlimit: 7
+    auto_bad: null
+    auto_bad_reject: null
+    auto_bad_flat: null
+    auto_bad_eeg_thresh: 10
+    auto_bad_meg_thresh: 10
+  head_position_estimation:
```
drammock: Since this file will be used as a template by users of varying levels of background knowledge, I think it needs more scaffolding than it currently has.  It's easy enough to add in comments later, but we have the potential to make the file somewhat self-documenting by making the key names very clear.  WDYT?  This key `head_position_estimation` seems like a good example of that; can we do similarly for all keys, or would that be too much of a pain?

(note: even if we do spell out the keys clearly, some comments are probably warranted, e.g., for things like `n_jobs` we should say `# integer or "cuda"`)

mnefun/_mnefun.py
```diff
@@ -129,33 +129,92 @@ def unfreeze(self):
 class Params(Frozen):
     """Make a parameter structure for use with `do_processing`
 
-    This is technically a class, but it doesn't currently have any methods
-    other than init.
+    Some attributes can be set on init, but it's better to use
```
drammock: better to use... ?

examples/funloc/funloc_params.yml
drammock: It would be nice if there were a way to only have to document in one place (here in the YAML?) and not also in the docstring of `Params`.  But if we do have to document in two places, I guess we could make the documentation in the YAML more sparse (i.e., limited just to the allowed values?)

examples/funloc/analysis_fun.py
larsoner: The analysis YAML is for a given project, how to process the data. It should be shareable by each person in the project, and its values should be specific to the given project.

The `mnefun.json` is a for a given machine/workstation, how to do SSS. It shouldn't be shareable across people, but its values should work for any project.

examples/funloc/funloc_params.yml
larsoner: > can we do similarly for all keys, or would that be too much of a pain?

The main category names (epoching, etc.) we can change. The other ones would be a pain to change (esp. in a backward compatible way) because they map onto the `Params` class attributes.

drammock: > The other ones would be a pain to change

OK, that's kinda what I thought

drammock:
> or maybe the best place to put real descriptions is in the Overview section of the docs with the flow chart.

I'm leaning toward this option actually.  It frees us from the constraints of docstrings, and avoids making the YAML file 80% comments and 20% actual param values.

larsoner:
Yeah I was thinking the same. It gives us the most flexibility. It also makes that page where people should go if they are confused about things.

larsoner:
@drammock unified in: https://51-17719360-gh.circle-artifacts.com/0/html/overview.html

drammock:
> @drammock unified in: https://51-17719360-gh.circle-artifacts.com/0/html/overview.html

beautiful!

larsoner:
Okay added `params.save` and YAML round-trip checking. I'll merge once CIs are happy, then @drammock @nordme hopefully you can try changing PreK or some other project over to YAML. You can hopefully do this by:

1. being on this branch (or soon latest `master`)
2. running your script with all steps disabled
3. doing `params.save('whatever.yml`)`
4. reconfigure your script to do `params = read_params('whatever.yml')`
5. inspect the `whatever.yml` to make sure it looks reasonable -- it should only contain params that differ from the defaults (but no harm in adding back ones that are defaults if you prefer to be explicit)

There might be some bugs but I want to merge this and iterate more later, that way you don't have to try this ASAP or anything. And in the meantime, I can move on to reorganizing the code into submodules, since the YAML should more or less guide the reorganization.

mnefun/data/canonical.yml
```diff
@@ -0,0 +1,179 @@
...
+  dates:
+  acq_ssh:
+  acq_dir:
+  acq_port:
+  run_names:
+  runs_empty:
+  subject_run_indices:
+
+scoring:
+  score:
```
ktavabi, line 19: How do I evoke `score.py` from here?

mnefun/data/canonical.yml
```diff
@@ -0,0 +1,179 @@
...
+    hp_cut:
+    hp_trans:
+    lp_cut:
+    lp_trans:
+    filter_length:
+    fir_design:
+    fir_window:
+    phase:
+  ssp:
+    proj_nums:
```
ktavabi, line 70: As is `params = mnefun.read_params(...)` doesn't bring up relevant `docstring` for e.g., order of projections by channel type

mnefun/data/canonical.yml
larsoner: This is one of the few things that we haven't yamlized, because there isn't a great way to do it. I'd just keep it in Python for now like:

https://github.com/LABSN/mnefun/blob/master/examples/funloc/analysis_fun.py#L31-L34

mnefun/data/canonical.yml
larsoner: Indeed currently all docs for `mnefun` params are in https://labsn.github.io/mnefun/overview.html

PR #272, opened 2019-11-27
MAINT: Refactor code into separate files

larsoner:
- [x] Refactor code into submodules to roughly match processing steps
- [x] Ensure tests pass locally
- [x] Ensure tests pass on CIs
- [x] Delete `subj_01`/`subj_02` locally (then `git checkout subj_01/*` and `02/*` to get prebads back)
- [x] Re-run `analysis_fun.py` with `default=True` to see if everything comes back


larsoner:
FYI @drammock I'm planning just to merge this once `analysis_fun.py` runs okay. In theory this shouldn't change any functionality, just make the code organization better. We can rename files and so forth later if it seems to make sense to do so.

PR #278, opened 2020-02-11
MRG: Add acquisition QA and Python head pos estimation

larsoner:
Need to test on `kasga` that it actually works. Todo:

- [x] Get write permissions on some machine with access to Brainstudio (drake?)
- [x] Install `miniconda` (if necessary)
- [x] Create `acq_qa` environment
- [x] Activate env, test it using `screen`
- [x] Add head position estimation (almost free after cHPI coil SNR)
- [x] Add python head position estimation mode
- [x] Ensure it works on empty-room files
- [x] Finish mnefun step documentation (head pos, coils, annots)
- [x] Run on some directories actually being acquired currently, check that outputs are good/useful


larsoner:
@nordme @NeuroLaunch can you look at the [updated workflow page](https://100-17719360-gh.circle-artifacts.com/0/html/overview.html) and see if it makes (some) sense to you? It's okay if it's not totally clear, just looking for incremental progress over the [current version](https://labsn.github.io/mnefun/overview.html), which is less complete.

Assuming it looks okay I'll go ahead and merge, then people can try out `p.hp_type = 'python'` (once they are on latest `master` branch of this and MNE).

NeuroLaunch:
Sure thing - I'll look over it soon.

larsoner:
@NeuroLaunch one thing to look at in particular is the updated instructions for setting up for MaxFilter use, can you see if the following helps you get your machine configured (or check that it's configured properly, or would have helped you set things up in the first place if you've done it already):

https://102-17719360-gh.circle-artifacts.com/0/html/overview.html#preparing-your-machine-for-maxwell-autobad-and-head-position-estimation

doc/overview.rst
```diff
@@ -41,11 +31,47 @@ when running on your machine::
...
 Using per-machine config files rather than per-script variables should
 help increase portability of scripts without hurting reproducibility
 (assuming we all use the same version of MaxFilter, which should be a
 safe assumption).
 
+To test that things are configured correctly, you can do:
+
+.. code-block:: console
+
+    $ python -c "import mnefun; mnefun.check_sws()"
```
NeuroLaunch: I verified that I could get this message (using a Jupyter notebook for documentation in my case).  I also was able to run the SSS step (set for local processing of tSSS and head position estimation on Kasga) on a subject that had previously been successful using the old way to set the sws parameters.

mnefun/_sss.py
```diff
@@ -1024,3 +1106,20 @@ def annotate_head_pos(raw, head_pos, rotation_limit=45, translation_limit=0.1,
 
     # Annotate on distance from the sensors
     return annot
+
+
+def check_sws():
```
NeuroLaunch: I see this new function creates a new instance of Params, which is fine because the sws parameters are populated from the config file in  Params.__init__().  But wouldn't the function have greater utility if it (optionally) accepted an existing 'params'?  Especially if a user for some reason has overwritten the parameter values after instantiation.  I tried this and it worked as expected: `check_sws(params)`

mnefun/_paths.py
```diff
@@ -242,3 +242,7 @@ def _prebad(p, subj):
 def _is_dir(d):
     """Safely check for a directory (allowing symlinks)"""
     return op.isdir(op.abspath(d))
+
+
+def _get_config_file():
+    return op.expanduser(op.join('~', '.mnefun', 'mnefun.json'))
```
NeuroLaunch: A user-set environmental variable with the name of JSON file is another way to go, but I see that the ".mnefun" directory is analogous to the ".mne/" directory already specified for MNE-Python.

mnefun/_sss.py
```diff
@@ -1024,3 +1106,20 @@ def annotate_head_pos(raw, head_pos, rotation_limit=45, translation_limit=0.1,
...
+    """Check if SSS workstation is configured correctly."""
+    from . import Params
+    p = Params()  # loads config
+    if p.sws_ssh is None:
+        raise RuntimeError('sws_ssh was not defined in mnefun config file %s'
+                           % (_get_config_file(),))
+    output = run_sss_command(None, '-version', None, host=p.sws_ssh,
```
NeuroLaunch: If I'm interpreting the code correctly, it is invoking a throw-away call to maxfilter (on the remote server) by just asking for the version.  I didn't initially have the maxfilter path set up, though in the new documentation you nicely request the user to make sure it is.  I haven't yet used maxfilter that way, as I've only used the remote server command for the head position estimation, and other future users may be in the same situation.  Might a call to a more universal remote bash command work, some sort of simple handshaking, or is there a reason to use an MNE-specific command?

mnefun/_sss.py
```diff
@@ -1024,3 +1106,20 @@ def annotate_head_pos(raw, head_pos, rotation_limit=45, translation_limit=0.1,
 
     # Annotate on distance from the sensors
     return annot
+
+
+def check_sws():
+    """Check if SSS workstation is configured correctly."""
+    from . import Params
+    p = Params()  # loads config
```
NeuroLaunch: I also noticed, even after getting everything else set up right, the first call to ```Params()``` to create an instance of 'params' always leads to the same warning: "UserWarning: Configuration value p.sws_dir was None, remote Maxfilter processing will fail."  etc for the other sws parameters.  Nevertheless, examination of the resulting 'params' shows that those parameters are correctly populated from the config file.

mnefun/_sss.py
drammock: Maybe I'm misunderstanding your point, but If the test just used a more universal bash command like `$ pwd`, then it wouldn't be checking/confirming that maxfilter is properly installed. So actually running `maxfilter -version` seems preferable to me, anyway.

NeuroLaunch:
I initially thought this function was designed to check that the server
info was valid for any processing (like head position without sss).  Now I
see its purpose is to specifically check for maxfilter on that server
(which I tested and it does that fine).  Unless a more general test is
needed, this part is ok.

doc/overview.rst
```diff
@@ -179,8 +233,10 @@ translation_limit : float
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 movecomp : str | None
     Movement compensation to use. Can be 'inter' or None.
+hp_type : str
```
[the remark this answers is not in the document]
larsoner: You can try already on this branch if you want

mnefun/_sss.py
larsoner: The idea is that everyone should start using the new way of doing things (i.e., machine-based config without values set in the processing script), so *not* allowing an existing `Params` is actually preferred here.

mnefun/_paths.py
larsoner: We can do it someday if we need to but for now I'd assume YAGNI

mnefun/_sss.py
larsoner: Indeed the idea is to not just make sure that SSH works, but also that the commands we're going to want to run will work (specifically, MaxFilter)

mnefun/_sss.py
larsoner: Whoops, this was a bug and should be fixed now

larsoner:
Thanks for the feedback @NeuroLaunch ! I am pushing (GitHub is having issues) a commit that addressed most of your concerns (I hope). I will probably just merge this as-is, but please do open a PR for further clarifications once you look over the changes.

> For fetch_raw, it might be worth noting that many users (Erica, at least, and now me) prefer to manually populate a local folder with the original raw data, skipping this step altogether.

I would recommend not doing this if possible, it indicates that something is wrong and can lead to inconsistencies. In theory you should not need to do this, unless things go wrong during acquisition (e.g., files being mis-named). I've written this up in the `overview` now, feel free to take a look.

PR #285, opened 2020-03-23
FIX: Apply projs before PSD

larsoner:


larsoner:
Merging since this fixes a definite problem and also fixes CIs

PR #286, opened 2020-04-02
MRG, ENH: prebad to YAML, plus defaultdict

larsoner:
@drammock this should simplify things in two ways:

1. It allows specifying pre-SSS bad channels directly in YAML using `mf_prebad` instead of using `_prebad.txt`
2. It allows a special `__default__` entry in YAML that causes the entry to become a `defaultdict` with the given entry as default. This is useful for example when most subjects have a specific set of values, but they need to be set to something different for a few others.

WDYT? Ready for review/merge from my end if you agree this is worthwhile.


larsoner:
Updated:

- [note about `'__default__'`](https://163-17719360-gh.circle-artifacts.com/0/html/overview.html#general-general-options)
- [flow chart](https://163-17719360-gh.circle-artifacts.com/0/html/overview.html#flow-chart)
- [parameter entry](https://163-17719360-gh.circle-artifacts.com/0/html/overview.html#preprocessing-pre-sss-bads-automatic-bad-channel-detection)

drammock:
Sorry to be slow to review this.  Maybe I'm not understanding correctly, but I think this actually won't work for the PREK data because it is longitudinal, and there's no guarantee that a given subject has the same `prebad` for both recording sessions. In fact, the first subject that I spot-checked indeed had different prebad channels for the two recordings.

drammock:
That said, the changes are still useful for non-longitudinal studies, so I'm +1 for merge if you'd rather not tackle the longitudinal problem right now.

doc/overview.rst
```diff
@@ -65,6 +65,12 @@ subject_indices : list
 disp_files : bool
     Display status.
 
+.. note:: Anywhere a ``dict`` is supported as an option (e.g.,
+          ``mf_prebad`` or ``proj_nums``), a special entry ``'__default__'``
+          can be used turn the dictionary into a
+          :class:`~python:collections.defaultdict` instance.
+          This is useful in cases where a single set of values works for most
+          subjects, but a few need different ones.
```
drammock, line 73: I think here it could be clearer that the special entry should be `__default__: [MEG8888, MEG9999]` or similar.

doc/overview.rst
```diff
@@ -153,6 +161,9 @@ n_jobs_resample : int | str
 ``preprocessing: pre-SSS bads``: Automatic bad channel detection
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+mf_prebad : dict
+    Dict with subject keys, with each value being a list of int of bad
+    MEG channels (e.g., ``[121, 1743]``).
```
drammock, line 166: oh, this is different from the prebad text file, right?  Where you said `MEG121` not just `121`. Or am I reading this wrong?

doc/overview.rst
drammock: EDIT: I see now it actually should be `__default__: [8888, 9999]` (without the "MEG" prefix)

larsoner:
> Maybe I'm not understanding correctly, but I think this actually won't work for the PREK data because it is longitudinal, and there's no guarantee that a given subject has the same prebad for both recording sessions. In fact, the first subject that I spot-checked indeed had different prebad channels for the two recordings.

This is the case in `master`, too -- the idea of this PR is just to make it so that instead of needing `_prebad.txt` files, you can define them in YAML. The issue of per-recording `_prebad.txt` or YAML entries we should probably address separately.

For example, if we force people to do per-recording `prebad` in YAML only (no support for it in files), then we can pretty easily allow the YAML entry to be either a list of str (current, global way) or list of list of str (what you need, a list per file).

doc/overview.rst
larsoner: Actually maybe this is a good time to abandon this number format. It's partly to save work with MaxFilter (which is being replaced anyway), but also partly to save typing `MEG` over and over again. But if you think it's cleaner to define them all as actual channel names like `['MEG0121', 'MEG1743']` instead of `[121, 1743]` we can do that here.

drammock: should probably ask @nordme and @mdclarke whether they prefer `121` vs `MEG121` here. I don't have a strong opinion.

My only maybe relevant thought is if mnefun were ever used with a different MEG manufacturer with different channel name conventions, would that require mnefun code changes to work. If we can avoid that by how we act now, it seems good to me.

drammock:
> > Maybe I'm not understanding correctly, but I think this actually won't work for the PREK data because it is longitudinal, and there's no guarantee that a given subject has the same prebad for both recording sessions. In fact, the first subject that I spot-checked indeed had different prebad channels for the two recordings.
> 
> This is the case in `master`, too -- the idea of this PR is just to make it so that instead of needing `_prebad.txt` files, you can define them in YAML. The issue of per-recording `_prebad.txt` or YAML entries we should probably address separately.
> 
> For example, if we force people to do per-recording `prebad` in YAML only (no support for it in files), then we can pretty easily allow the YAML entry to be either a list of str (current, global way) or list of list of str (what you need, a list per file).

Ah, I guess you're right, right now we effectively have 2 separate mnefun scripts (really, the same script with a flag changed) for our two longitudinal timepoints. So yeah, I guess this is an orthogonal issue.

larsoner:
> really, the same script with a flag changed

Is this something else that we could unify so that you could have a single script?

drammock:
> > really, the same script with a flag changed
> 
> Is this something else that we could unify so that you could have a single script?

Maybe. I'll give it some thought.  At present the first and second recordings for each subj are in completely separate directory trees, but it would be nice to let them coexist.

doc/overview.rst
mdclarke: I think using `MEG` makes sense to stay consistent and for the reason Dan listed above.

larsoner:
Okay, changed the format to be actual channel names and updated the doc. I'll go ahead and merge and feel free to try it out and report bugs!

drammock:
thx @larsoner!

PR #291, opened 2020-04-24
FIX: split-file issue with epoching and covariance processing (closes 287)

NeuroLaunch:
I tested this with 1) the usual funloc examples with non-split raw files and only a single run (subj_01 and subj_02); 2) a pseudo "subj_03" (also non-split), with a second run created by duplicating and clipping the original raw fif data; 3) a f2f subject with split raw files (2 file parts) encompassing a single run.  Note, I did not test it on a subject with split raw files AND multiple runs, but the code should allow for it.

Also corrected minor issues in other files, but please see commit notes.

### DIFF
--- a/mnefun/_scoring.py
+++ b/mnefun/_scoring.py
@@ -115,8 +115,8 @@ def _read_events(p, subj, ridx, raw):
                                % (fname,))
         events.append(these_events)
     if len(events) == 1 and len(raw._first_samps) > 1:  # for split raw
-        first_samps = np.array(raw._first_samps[0], ndmin=1)
-        last_samps = np.array(raw._last_samps[-1], ndmin=1)
+        first_samps = raw._first_samps[:1]
+        last_samps = raw._last_samps[-1:]
     else:
         first_samps = raw._first_samps
         last_samps = raw._last_samps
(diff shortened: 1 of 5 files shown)


mnefun/_scoring.py
```diff
@@ -114,7 +114,13 @@ def _read_events(p, subj, ridx, raw):
             raise RuntimeError('Non-unique event samples found in %s'
                                % (fname,))
         events.append(these_events)
-    events = concatenate_events(events, raw._first_samps, raw._last_samps)
+    if len(events) == 1 and len(raw._first_samps) > 1:  # for split raw
+        first_samps = np.array(raw._first_samps[0], ndmin=1)
```
larsoner, line 118: I've more often seen `np.atleast_1d` for these sorts of things, but here I would do something more compact:
```
first_samps = raw._first_samps[:1]
last_samps = raw_last_samps[-1:]
```

NeuroLaunch: .atleast_1d() is new to me, but the slice trick is golden.  I'll fix that and let you merge.  Thank you.

larsoner:
Thanks @NeuroLaunch !

PR #292, opened 2020-04-28
ENH: Add every_other

### LINKED ISSUE #288: ENH: Every-other averages added to `-evo.fif`
Samu has been asking about this for a while, I realized we can just add it to the `-ave.fif` files that we generate.

One open question would be how to add these to the reports. For a first pass we don't automatically do it and just let people add them to their configs.

larsoner:
1. Forces `cov` to be provided when `peaks='times'` because the `rank` is needed to compute the GFP. I don't think this is too onerous anyway...
2. Adds `every_other` support, plus plotting in `report` if they're available in the Evoked

@NeuroLaunch can you try this on Alexis's data?

Closes #288

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -454,6 +454,10 @@ must_match : list of int
     Indices from the original in_names that must match in event counts
     before collapsing. Should eventually be expanded to allow for
     ratio-based collapsing.
+every_other : bool
+    If True, in addition to standard averages / evoked data, averages will be
+    computed from every other trial, i.e., from even and odd trials separately.
+    This can help assess the SNR of the data.
 
 8. gen_covs
 -----------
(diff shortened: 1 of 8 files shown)


NeuroLaunch:
Yes I can try it out.  I think she's intending it for GenZ?  I'll ask Alexis for a good data subject to try.

NeuroLaunch:
The feature looks ok, with funloc and a sample f2f subject.  I will also try it with a genZ subject with a pared-down epoch creation, with an eye on bugs rather than meaningful analysis - I'll let you know if any problems come up with that.

I like your exploit of the _mne.read_evokeds()_ name/condition parameter to read in the odd and even data, as using the 'kind' parameter is restrictive.  I used that to read in average and std error waveforms separately, then plotted odd and even averages, side-by-side, with error indicators.

I also like the simplicity for the reports that if odd/even potentials are available it just includes them.  I'd suggest using the same y-axis limits for the standard, odd, and even plots (per respective channel types), if it's not much trouble.

PR #293, opened 2020-05-07
MRG, FIX: Rank

larsoner:
Requires some upstream changes to MNE to handle `compute_rank` properly. Going to keep this open while debugging GenZ cov issues.

Also moves toward making MNE_3D_BACKEND=pyvista usable with reports.


larsoner:
Going to merge and continue with report changes in a subsequent PR. @nordme with this and latest MNE master you should be able to regenerate covariances + invs once I send you the updated script

PR #295, opened 2020-06-19
MRG, FIX: Fix errant picking of only data channels

larsoner:
FYI @NeuroLaunch if you rerun the `apply_ssp` and `write_epochs` steps, your `-epo.fif` and `-ave.fif` files should end up with the cHPI channels included. At least with the `funloc` data these changes make it work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,11 @@ env:
   global: OPENBLAS_NUM_THREADS=1
 
 python:
-  - "3.7"
+  - "3.8"
 
 install:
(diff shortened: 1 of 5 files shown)


NeuroLaunch:
> FYI @NeuroLaunch if you rerun the `apply_ssp` and `write_epochs` steps, your `-epo.fif` and `-ave.fif` files should end up with the cHPI channels included. At least with the `funloc` data these changes make it work.

I'll try that out - thank you!

PR #298, opened 2020-07-08
MRG, ENH: Add report hooks

larsoner:
Allow adding additional entries to the report at the beginning and end:

Screenshot from 2020-07-08 15-54-22

Screenshot from 2020-07-08 15-54-26

@NeuroLaunch will this help with F2F? This is a super simple example but in principle you can use any of the `report` functions you want, including the `add_figs_to_section`, etc.


NeuroLaunch:
This seems right what I was hoping for as a report feature.  Let me take a closer look, and then I'll test it out on a branch.  Thanks!

larsoner:
Added a commit to give the ERMs and processed ERMs in the report, as well as a "proj off" option in the `report_params['sensor']`. Still need to add the reconstruction step...

PR #301, opened 2020-08-20
FIX: Adjust coil time window to allow for line harmonics

NeuroLaunch:
This pull request fixes an SSS processing issue with poor coil amplitude fits (goodness-of-fits << 0.90) when coil frequencies are too close to harmonics of the line frequency.
- Removed a redundant `t_window` calculation in _head_pos_annot()
- Added option to return `t_window` value from call to _get_fit_data(), for backward compatibility
- Fixed an occasional problem where raw.time_as_index() returns a negative value other than -1 (the mne-python method shouldn't do this, but the fix here in mnefun should not cause conflicts)
- Caveat: The `t_window` value used for filter_chpi() and other routines may not be the same as `t_window` calculated for _get_fit_data(); this seems not be a problem
- Code was tested with 1 subject having high coil frequencies (700-900+ Hz) and 1 subject having more typical frequencies (83-323 Hz).

### DIFF
--- a/mnefun/_sss.py
+++ b/mnefun/_sss.py
@@ -582,7 +582,7 @@ def _get_t_window(p, raw):
             info = raw.info
             if info['line_freq'] is not None:  # as in mne.chpi.py
                 line_freqs = np.arange(info['line_freq'], info['sfreq'] / 3.,
-                       info['line_freq'])
+                                       info['line_freq'])
             else:
                 line_freqs = np.array([60])   # SMB: should be safe to assume
         except RuntimeError:
@@ -594,7 +594,7 @@ def _get_t_window(p, raw):
             # of 727 Hz+) and 83 ms for 60 Hz line freq (w/ more typical coil
             # freqs of 83 Hz+).
             all_freqs = np.concatenate((hpi_freqs, line_freqs))
-            delta_freqs = [abs(b-a) for b in all_freqs for a in all_freqs \
+            delta_freqs = [abs(b - a) for b in all_freqs for a in all_freqs
                            if a != b]
             delta_freqs = np.array(delta_freqs)
             t_window = max(5. / all_freqs.min(), 1. / delta_freqs.min())
--- a/mnefun/_sss.py
+++ b/mnefun/_sss.py
@@ -16,7 +16,7 @@ from scipy.spatial.distance import cdist
 from mne import (read_trans, Transform, read_annotations, Annotations,
                  pick_types)
 from mne.bem import fit_sphere_to_headshape
-from mne.chpi import read_head_pos, write_head_pos, filter_chpi
+from mne.chpi import read_head_pos, write_head_pos, filter_chpi, _get_hpi_info
 from mne.externals.h5io import read_hdf5, write_hdf5
 from mne.io import read_raw_fif, BaseRaw, write_info, read_info
 from mne.preprocessing import maxwell_filter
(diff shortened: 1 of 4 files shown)


mnefun/_sss.py
```diff
@@ -577,16 +577,27 @@ def _pos_valid(pos, dist_limit, gof_limit):
...
     if t_window == 'auto':
-        from mne.chpi import _get_hpi_info
         try:
             hpi_freqs, _, _ = _get_hpi_info(raw.info, verbose=False)
+            info = raw.info
+            if info['line_freq'] is not None:  # as in mne.chpi.py
+                line_freqs = np.arange(info['line_freq'], info['sfreq'] / 3.,
+                                       info['line_freq'])
+            else:
+                line_freqs = np.array([60])   # SMB: should be safe to assume
```
larsoner, line 587:
```suggestion
            info = raw.info
            line_freq = info['line_freq'] or 60.
            line_freqs = np.arange(line_freq, info['sfreq'] / 3., line_freq)
```
the `or` short-circuit trick here should be safe enough for us

mnefun/_sss.py
```diff
@@ -577,16 +577,27 @@ def _pos_valid(pos, dist_limit, gof_limit):
...
-            t_window = max(5. / min(hpi_freqs), 1. / np.diff(hpi_freqs).min())
+            # Use the longer of 5 cycles and the difference across all
+            # HPI and line harmonic frequencies.
+            # This will be 143 ms for 7 Hz spacing (e.g. w/ higher coil freqs
+            # of 727 Hz+) and 83 ms for 60 Hz line freq (w/ more typical coil
+            # freqs of 83 Hz+).
+            all_freqs = np.concatenate((hpi_freqs, line_freqs))
+            delta_freqs = [abs(b - a) for b in all_freqs for a in all_freqs
+                           if a != b]
+            delta_freqs = np.array(delta_freqs)
```
larsoner, line 599:
```suggestion
            delta_freqs = np.diff(np.unique(np.concatenate((hpi_freqs, line_freqs))))
```

mnefun/_sss.py
```diff
@@ -696,7 +707,7 @@ def _mf_autobad(raw, p, skip_start, skip_stop):
     return bads
 
 
-def _get_fit_data(raw, p=None, prefix='    '):
+def _get_fit_data(raw, p=None, prefix='    ', return_t_window=False):
```
larsoner, line 710: rather than add a parameter, this func is private so we can just always return it and update all uses of `_get_fit_data` in `mnefun`

NeuroLaunch:
I'll make some comments on whats_new and make the other suggested changes.
Thanks.

mnefun/_sss.py
NeuroLaunch: That's a nice simplification, because np.unique() puts the frequencies in order (so larger differences don't even have to be computed).  Thanks!

mnefun/_sss.py
```diff
@@ -577,16 +577,22 @@ def _pos_valid(pos, dist_limit, gof_limit):
 def _get_t_window(p, raw):
     t_window = p.coil_t_window if p is not None else 'auto'
     if t_window == 'auto':
-        from mne.chpi import _get_hpi_info
         try:
             hpi_freqs, _, _ = _get_hpi_info(raw.info, verbose=False)
+            info = raw.info
+            line_freq = info['line_freq'] or 60
+            line_freqs = np.arange(line_freq, info['sfreq'] / 3., line_freq)
```
larsoner, line 584: In general you want as few as possible commands in the `try` part as possible. In this case the `try` has to do with `_get_hpi_info` failing on old files IIRC, so your new lines of code should all live in the `else` clause

mnefun/_sss.py
```diff
@@ -779,9 +785,9 @@ def _head_pos_annot(p, subj, raw_fname, prefix='  '):
     if p is not None and p.movecomp is None:
         head_pos = fit_data = None
     else:
-        t_window = _get_t_window(p, raw)
+        # t_window = _get_t_window(p, raw)    SMB 20.08.20: incorp. below
```
larsoner, line 788: you can just delete this line, it will live on in `git` history/blame if we ever need it again. Nice that these changes make the code more DRY !

NeuroLaunch: The one extra fix I made here, just for accounting purposes, is to set `t_window` to _None_ along with the other parameters.  It's not required unless the subsequent `if` clause is satisfied, but I believe it's good practice if an early `return` is not explicitly declared.

larsoner:
Thanks @NeuroLaunch ! Can you make a PR to MNE to update the corresponding function:

I think the difference there will be that if `info.get('line_freq', None) is None`, then you don't include the harmonics because in that case we can't assume 60 vs 50 Hz line noise (like we can for UW data)

larsoner:
https://github.com/mne-tools/mne-python/blob/master/mne/chpi.py#L450

NeuroLaunch:
It's on my to-do list.  Thanks for the tip about the 60 Hz line frequency.

PR #310, opened 2020-09-30
ENH: Add options for HEOG/VEOG

larsoner:
@nordme feel free to give this a try, hopefully it makes things flexible enough!


larsoner:
(I did not test it locally, though, so if you hit an error feel free to report it here quickly rather than trying to debug things)

PR #315, opened 2020-12-04
proportion → number

drammock:
The docs say "proportion" but what is actually compared is just the raw number, as far as I can tell from the code here:

https://github.com/LABSN/mnefun/blob/4885000a57fb61f0ea3cd01e65dcd327faf557f5/mnefun/_ssp.py#L200-L208

I think this is the right fix given that the default is "10" (i.e., not a proportion), but LMK if you'd rather I leave the docs alone and fix the code / default.


larsoner:
Thanks @drammock

PR #316, opened 2020-12-18
ENH: Allow float decim as dest

larsoner:
Useful for @ktavabi who has data sampled at 1800 and 1200 Hz, allowing `decim=300.` is a lot nicer than having to pass a list of decim values. @nordme @mdclarke can you look as well, as it might be relevant to you, too? Some of your data might be this way, not sure.

### DIFF
--- a/mnefun/_utils.py
+++ b/mnefun/_utils.py
@@ -80,6 +80,19 @@ def _handle_dict(entry, subj):
     return out
 
 
+def _handle_decim(decim, sfreq):
+    decim = np.array(decim)
+    assert decim.shape == ()
+    if decim.dtype.char in 'il':
+        return decim
+    else:
+        # float
+        assert decim.dtype.char == 'd', decim.dtype.char
+        got_decim = int(round(sfreq / decim))
+        assert np.isclose(sfreq / got_decim, decim), (sfreq, decim, got_decim)
+        return got_decim
+
+
 def _safe_remove(fnames):
     if isinstance(fnames, str):
         fnames = [fnames]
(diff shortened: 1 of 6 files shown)


mdclarke:
This works for me. Thanks for this, it's helpful for a lot of datasets :)

larsoner:
Okay great, in it goes! Others still feel free to look and comment

mnefun/_utils.py
```diff
@@ -80,6 +80,19 @@ def _handle_dict(entry, subj):
     return out
 
 
+def _handle_decim(decim, sfreq):
+    decim = np.array(decim)
+    assert decim.shape == ()
```
ktavabi, line 85: I assume you tested and this works, but what's the reasoning for this assertion? Isn't always going to be a 'tuple'..is it like a stealthy `type` check?

mnefun/_utils.py
```diff
@@ -80,6 +80,19 @@ def _handle_dict(entry, subj):
...
+    if decim.dtype.char in 'il':
```
ktavabi, line 86: Again, for my own personal edification: what is `il`?

mnefun/_utils.py
larsoner: `np.array(..)` can return arbitrary shapes, this ensures it's a scalar as:
```
>>> np.array(1.).shape
()
>>> np.array([1.]).shape
(1,)
```
etc.

mnefun/_utils.py
larsoner: int or long

ktavabi: Maybe comment w/ _NumPy dtype attribute for workspace names_ if that's the correct terminology. That's some Kung Fu Panda shit :)

ktavabi:
Feel free to close.

PR #317, opened 2020-12-18
ENH: Add eSSS

larsoner:
Adds `eSSS` with optional band-pass support to `mnefun`. If `params.erm_proj_as_ess = True` then it will use `proj_nums[2]` for a given subject to compute empty-room projectors after band-pass filtering between `params.erm_proj_hp_cut` and `params.erm_proj_lp_cut`. You'll probably need to set `params.erm_proj_reject = {'mag': 5.e-11, 'grad': 1.5e-9}` or something similar since the empty-room data can be pretty noisy.

@NeuroLaunch @nordme @mdclarke (I can't remember who all had the issue) feel free to try this to see if it helps with any 30 Hz frequency bumps you see. I'm not 100% convinced that the bad channels are handled properly here, but basically it uses `raw_fif/subj_prebad.txt` to choose bad channels as these will be marked bad for *all* files that will be processed by SSS. So sometimes it might be important to make sure this is marked properly otherwise you might end up with a single channel dominating.

In any case, it doesn't seem to make a huge difference in the 30 Hz bump for the `funloc` data even with 8 projectors, but the bump wasn't huge there to begin with:

Screenshot from 2020-12-18 15-45-39

It's possible there is some bug -- @NeuroLaunch please try it on some dataset with a prominent peak and see if it makes it at least go down a little bit. That would suggest it's at least implemented somewhat properly. And then you could compare re-running with `params.erm_proj_as_esss = False` just as a basis for comparison. For these comparisons I think you'll have to re-run from SSS, computing SSP, applying SSP, epoching, cov, inv, and reports. I will not have time to push fixes for a while so feel free to make a PR into my branch or start a new PR with my commits if you want.

### DIFF
--- a/examples/funloc/funloc_params.yml
+++ b/examples/funloc/funloc_params.yml
@@ -44,7 +44,7 @@ preprocessing:
     tsss_dur: 60.
     st_correlation: 0.98
     trans_to: twa
-    erm_proj_as_esss: True
+    cont_as_esss: True
   filtering:
     lp_cut: 50
     lp_trans: 10
@@ -52,13 +52,14 @@ preprocessing:
     fir_design: firwin
     phase: zero
   ssp:
+    cont_hp: 20
+    cont_hp_trans: 2
+    cont_lp: 40
+    cont_lp_trans: 2
     proj_nums: {
-      subj_01: [[2, 2, 0], [1, 1, 3], [8, 8, 0], [1, 1, 0], [1, 1, 0]],
-      subj_02: [[2, 2, 0], [1, 1, 2], [8, 8, 0]],
+      subj_01: [[2, 2, 0], [1, 1, 3], [3, 3, 0], [1, 1, 0], [1, 1, 0]],
+      subj_02: [[2, 2, 0], [1, 1, 2], [3, 3, 0]],
       }
-    erm_proj_lp_cut: 20
-    erm_proj_lp_cut: 40
-    erm_proj_reject: {'mag': 5.e-11, 'grad': 1.5e-9}
     proj_sfreq: 200
     proj_meg: combined
     eog_f_lims: [1, 10]
@@ -108,7 +109,7 @@ report:
   sensor: [
     {analysis: All, name: All, times: 'peaks', proj: False},
     {analysis: All, name: All, times: 'peaks'},
-    {analysis: All, name: All, times: 'peaks', proj: 'reconstruct'},
+    #{analysis: All, name: All, times: 'peaks', proj: 'reconstruct'},
     ]
   source: {analysis: All, name: All, inv: '%s-50-sss-meg-eeg-free-inv.fif',
            times: 'peaks', views: lat, size: [800, 400]}
--- a/examples/funloc/funloc_params.yml
+++ b/examples/funloc/funloc_params.yml
@@ -44,6 +44,7 @@ preprocessing:
     tsss_dur: 60.
     st_correlation: 0.98
     trans_to: twa
+    erm_proj_as_esss: True
   filtering:
     lp_cut: 50
     lp_trans: 10
@@ -52,9 +53,12 @@ preprocessing:
     phase: zero
   ssp:
     proj_nums: {
-      subj_01: [[2, 2, 0], [1, 1, 3], [0, 0, 0], [1, 1, 0], [1, 1, 0]],
-      subj_02: [[2, 2, 0], [1, 1, 2], [0, 0, 0]],
+      subj_01: [[2, 2, 0], [1, 1, 3], [8, 8, 0], [1, 1, 0], [1, 1, 0]],
+      subj_02: [[2, 2, 0], [1, 1, 2], [8, 8, 0]],
       }
+    erm_proj_lp_cut: 20
+    erm_proj_lp_cut: 40
+    erm_proj_reject: {'mag': 5.e-11, 'grad': 1.5e-9}
     proj_sfreq: 200
     proj_meg: combined
     eog_f_lims: [1, 10]
--- a/examples/funloc/funloc_params.yml
+++ b/examples/funloc/funloc_params.yml
@@ -109,7 +109,7 @@ report:
   sensor: [
     {analysis: All, name: All, times: 'peaks', proj: False},
     {analysis: All, name: All, times: 'peaks'},
-    #{analysis: All, name: All, times: 'peaks', proj: 'reconstruct'},
+    {analysis: All, name: All, times: 'peaks', proj: 'reconstruct'},
     ]
   source: {analysis: All, name: All, inv: '%s-50-sss-meg-eeg-free-inv.fif',
            times: 'peaks', views: lat, size: [800, 400]}
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -267,6 +267,10 @@ class Params(Frozen):
         self.every_other = False
         self.cov_rank_method = 'estimate_rank'
         self.epochs_proj = True
+        self.erm_proj_as_esss = False
+        self.erm_proj_lp_cut = None
+        self.erm_proj_hp_cut = None
+        self.erm_proj_reject = None
         self.freeze()
         # Read static-able paraws from config file
         _set_static(self)
--- a/mnefun/_mnefun.py
+++ b/mnefun/_mnefun.py
@@ -110,7 +110,7 @@ class Params(Frozen):
         self.cont_hp = None
         self.cont_hp_trans = 0.5
         self.cont_lp = 5.
-        self.cont_lp_trans = None
+        self.cont_lp_trans = 0.5
         self.hp_cut = hp_cut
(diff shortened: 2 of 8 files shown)


examples/funloc/funloc_params.yml
```diff
@@ -52,9 +53,12 @@ preprocessing:
     phase: zero
   ssp:
     proj_nums: {
-      subj_01: [[2, 2, 0], [1, 1, 3], [0, 0, 0], [1, 1, 0], [1, 1, 0]],
-      subj_02: [[2, 2, 0], [1, 1, 2], [0, 0, 0]],
+      subj_01: [[2, 2, 0], [1, 1, 3], [8, 8, 0], [1, 1, 0], [1, 1, 0]],
+      subj_02: [[2, 2, 0], [1, 1, 2], [8, 8, 0]],
       }
+    erm_proj_lp_cut: 20
+    erm_proj_lp_cut: 40
```
NeuroLaunch, line 60: Probably meant `erm_proj_hp_cut: 20` and lp cut at 40.

larsoner: Yikes this made me realize that the code was wrong, and that the band-passing was almost certainly not doing what it was supposed to. I also reused `p.cont_lp` and `p.cont_hp` because they already existed. Feel free to look at the updated docs/params and give it a shot

NeuroLaunch: I plan to make a run through with a good subject example soon, and I'll let you know if I come across a bug or have questions about the algorithm (especially how the revised SSS step relates to subsequent SSP processing).

NeuroLaunch:
I tried the eSSS on subject f2f_107 and it seems to help lower the bump at 29 Hz and improve denoising overall.  I'll know more when I put more subjects through the full pipeline.

Here is original followed by with-eSSS for the ERM raw, with bandpass from 25 to 32 Hz and 1 Hz transition:
Screen Shot 2021-01-16 at 5 34 36 PM
Screen Shot 2021-01-16 at 5 34 55 PM

Here is original and with-eSSS for the data raw:
Screen Shot 2021-01-16 at 5 36 36 PM
Screen Shot 2021-01-16 at 5 36 18 PM

The post-processing peak of the original wasn't too big for this subject, but it is completely gone in the eSSS case.  eSSS may have introduced other higher-frequency features, so I don't know what to make of that.  Also, the 60 Hz notch is narrower in the original case, but I think that's because I had been using different parameters for one of the processing steps; I'd prefer the narrower notch, so if you can remind me what may be influencing that (p.coil_t_window?) please let me know.

mnefun/_mnefun.py
```diff
@@ -108,11 +108,13 @@ def __init__(self, tmin=None, tmax=None, t_adjust=0, bmin=-0.2, bmax=0.0,
         self.n_jobs_resample = n_jobs_resample
         self.filter_length = filter_length
         self.cont_hp = None
+        self.cont_hp_trans = 0.5
         self.cont_lp = 5.
-        self.lp_cut = lp_cut
+        self.cont_lp_trans = None
         self.hp_cut = hp_cut
-        self.lp_trans = lp_trans
+        self.lp_cut = lp_cut
```
NeuroLaunch, line 115: I got an error when `p.cont_lp_trans` was none, so I made sure it was set to a reasonable value.

mnefun/_sss.py
```diff
@@ -411,6 +412,19 @@ def run_sss_locally(p, subjects, run_indices):
...
+            assert p.proj_meg == 'combined', 'must be combined for eSSS'
+            proj_nums = _proj_nums(p, subj)[2]
+            if any(proj_nums):
+                prebad_file = (p, subj, dict(autobad=False, prefix=' ' * 8))
+                print('      Computing eSSS basis from ERM with %d components'
+                      % (proj_nums[0],))
+                kwargs['extended_proj'] = _compute_erm_proj(
+                    p, subj, projs=[], kind='raw', bad_file=prebad_file,
+                    remove_existing=True, disp_files=False)
+
```
NeuroLaunch, line 427: It would be nice here (maybe within _computer_erm_proj()) to have a printed confirmation of the projector file name saved (_prepoc_cont....fif_).

NeuroLaunch: Also, I was wrong about there being an error with bad channels (such that prebads were loaded from a .txt file but not via the yml parameters).  I believe I originally misinterpreted a section in __ssp.py_ involving **postbad** channels, which as you said are usually just for EEG channels.

NeuroLaunch:
Here's a list of additional changes I hope you can make in this PR that will help my MNEFun code read in certain event files (that contain multiple event types sharing a timestamp).  The changes are pretty simple (a few lines in 3 files) and shouldn't impact greatly existing pipelines.  I think, in fact, that the change may be positive in that only epochs related to the targeted processing will be considered** when evaluating covariance and making epochs for calculating rejection criteria.

The main change is in _scoring.py at _read_events():
`def _read_events(p, subj, ridx, raw):`
becomes
`def _read_events(p, subj, ridx, raw, ids=None):`  at line 107
and
```
        these_events = read_events(fname)
```
becomes
```
        these_events = read_events(fname, include=ids)
```
  
Lines that call this function are in_epoching.py at line 105:
`        events = _read_events(p, subj, run_indices[si], raw, ids=in_numbers)`
 and in_cov.py at line 146:
`            events = _read_events(p, subj, ridx, raw, ids=p.in_numbers)`

**If you think it's best that _ids=params.in_numbers_ should be optional then perhaps a new parameter like `params.restrict_events = True` can be implemented, with the default set to _False_.

These modifications of course have nothing to do with the eSSS changes, so I'd be happy to put them in a different pull request if you'd like.  In any case, that's all the input I have for this PR for now.  I'll be using eSSS on upcoming processing and will get back to you if it causes unexpected problems.

larsoner:
>  I'd prefer the narrower notch, so if you can remind me what may be influencing that (p.coil_t_window?) please let me know.

Yep, that's the one.

For the other unrelated comment I turned it into a separate issue that we should tackle in a separate PR as they seem unrelated to this one. Okay to merge this one then?

NeuroLaunch:
> > I'd prefer the narrower notch, so if you can remind me what may be influencing that (p.coil_t_window?) please let me know.
> 
> Yep, that's the one.
> 
> For the other unrelated comment I turned it into a separate issue that we should tackle in a separate PR as they seem unrelated to this one. Okay to merge this one then?

Yes, please go ahead.  Thanks!

mnefun/_mnefun.py
larsoner: Ahh thanks, fixed

mnefun/_sss.py
larsoner: > o have a printed confirmation of the projector file name saved (prepoc_cont....fif).

For this it's a standard path and most people most of the time shouldn't care, so I'd rather not add this to the printed outputs.

examples/funloc/funloc_params.yml
```diff
@@ -44,16 +44,21 @@ preprocessing:
...
     filter_length: auto
     fir_design: firwin
     phase: zero
   ssp:
+    cont_hp: 20
+    cont_hp_trans: 2
+    cont_lp: 40
+    cont_lp_trans: 2
     proj_nums: {
-      subj_01: [[2, 2, 0], [1, 1, 3], [0, 0, 0], [1, 1, 0], [1, 1, 0]],
```
ktavabi, line 55: Before noticing these lists I was warming up to the idea of `yaml` file(s). I assume if the `param` is not listed in the `yaml` I can proceed as before? @larsoner

NeuroLaunch: My experience is that @larsoner has been good about setting reasonable `params `defaults for all fields left undefined in the yaml file.  The yaml approach makes the code cleaner and (for me) a way to archive previous mnefun pipeline runs.

ktavabi: My only concern with `{k:v}` where `v = list([],[])` is that the yaml can get quite unruly for a run on 18-20 subjects.

NeuroLaunch: I see.  I haven't used the double lists on proj_nums myself, though my last run had 35 subjects.  Just keep track of those commas, I guess!

ktavabi: Actually the [MNE-BIDS API ](https://mne.tools/mne-bids/stable/generated/mne_bids.mark_bad_channels.html?highlight=bad%20channels) is built to comb through lists so this a useful approach for downstream integration. WRT to ACQ bad channels I think we should always point the to the `prebad.txt`, there is really no need to duplicate the contents of that SLF anywhere else.

larsoner: Yes YAML just overrides what are defaults in `Params`, which are certainly not optimal defaults but should be backward compatible.

Regarding bads personally if I were starting a study from scratch I wouldn't use any `_prebad.txt` anymore but would instead put everything in the YAML. It makes the definition longer but it's fewer files to commit and keep track of.

PR #323, opened 2021-02-17
FIX/ENH: Event picking for autoreject + epoching

### LINKED ISSUE #319: ENH: More event processing
Here's a list of additional changes I hope you can make in this PR that will help my MNEFun code read in certain event files (that contain multiple event types sharing a timestamp).  The changes are pretty simple (a few lines in 3 files) and shouldn't impact greatly existing pipelines.  I think, in fact, that the change may be positive in that only epochs related to the targeted processing will be considered** when evaluating covariance and making epochs for calculating rejection criteria.

The main change is in _scoring.py at _read_events():
`def _read_events(p, subj, ridx, raw):`
becomes
`def _read_events(p, subj, ridx, raw, ids=None):`  at line 107
and
```
        these_events = read_events(fname)
```
becomes
```
        these_events = read_events(fname, include=ids)
```
  
Lines that call this function are in_epoching.py at line 105:
`        events = _read_events(p, subj, run_indices[si], raw, ids=in_numbers)`
 and in_cov.py at line 146:
`            events = _read_events(p, subj, ridx, raw, ids=p.in_numbers)`

**If you think it's best that _ids=params.in_numbers_ should be optional then perhaps a new parameter like `params.restrict_events = True` can be implemented, with the default set to _False_.

These modifications of course have nothing to do with the eSSS changes, so I'd be happy to put them in a different pull request if you'd like.  In any case, that's all the input I have for this PR for now.  I'll be using eSSS on upcoming processing and will get back to you if it causes unexpected problems.

_Originally posted by @NeuroLaunch in https://github.com/LABSN/mnefun/issues/317#issuecomment-761938280_

NeuroLaunch:
Fixes #319 

Partly this PR fixes an error in _\_scoring.py:\_read_events()_ when two events of different ids have the same timestamp.  The simplest workaround would be to pull in only the events that are to be later processed.  At the suggestion of @larsoner I retained existing code for picking events for covariance estimation via a user-defined function, and ported a similar step for the auto_rejection step.  For convenience (e.g. to overcome the double-timestamp issue), the string "restrict" can be used instead of the user-defined picking function; this is interpreted simply as using the events already chosen for processing via `params.in_numbers`.  After the picking, an error is given if repeated timestamps are found.

I think this method gives the most flexibility for choosing events for epoching, depending on how those epochs will be used, whether it's for calculating covariance or rejection thresholds.  Note that an extra event reading step was added for the _final_ epoching (i.e. for epochs that will be written to disk); in this case the "restrict" option is used because only and all events marked for analysis should be included at this step.  (The "restrict" option should work even if `params.in_numbers` is `None`, because then `include=None` when executing _mne.read\_events()_.)

One potential point of failure is that, within _\_read_events()_, repeated timestamps are checked only _after_ events are concatenated across files. I had no practical way to check this situation (I left in the original code, commented out).

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -491,6 +491,9 @@ autoreject_types : tuple
     reject trials on basis of EOG.
 reject_epochs_by_annot : bool
     If True, reject epochs by BAD annotations.
+pick_events_autoreject : callable | string | None
+    Function for picking covariance events, or the string "restrict"
+    to limit events to those with an id in ``in_numbers``.
 analyses : list of str
     Lists of analyses of interest.
 in_names : list of str
@@ -526,8 +529,9 @@ cov_method : str
 compute_rank : bool
     Default is False. Set to True to compute rank of the noise covariance
     matrix during inverse kernel computation.
-pick_events_cov : callable | None
-    Function for picking covariance events.
+pick_events_cov : callable | string | None
+    Function for picking covariance events, or the string "restrict"
+    to limit events to those with an id in ``in_numbers``.
 cov_rank : str | int
     Cov rank to use, usually "auto".
 cov_rank_method : str
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -492,7 +492,7 @@ autoreject_types : tuple
 reject_epochs_by_annot : bool
     If True, reject epochs by BAD annotations.
 pick_events_autoreject : callable | string | None
-    Function for picking covariance events, or the string "restrict"
+    Function for picking autoreject events, or the string "restrict"
     to limit events to those with an id in ``in_numbers``.
 analyses : list of str
     Lists of analyses of interest.
--- a/mnefun/_cov.py
+++ b/mnefun/_cov.py
@@ -144,7 +144,7 @@ def gen_covariances(p, subjects, run_indices, decim):
             this_decim = _handle_decim(decim[si], raw.info['sfreq'])
             # read in events
             picker = p.pick_events_cov
-            if type(picker) is str:
+            if isinstance(picker, str):
                 assert picker == 'restrict', \
                     'Only "restrict" is a valid string for p.pick_events_cov'
             elif picker and not callable(picker):
--- a/mnefun/_cov.py
+++ b/mnefun/_cov.py
@@ -143,17 +143,13 @@ def gen_covariances(p, subjects, run_indices, decim):
             raw = concatenate_raws(raws)
             this_decim = _handle_decim(decim[si], raw.info['sfreq'])
             # read in events
-            events = _read_events(p, subj, ridx, raw)
-            if p.pick_events_cov is not None:
-                old_count = sum(len(e) for e in events)
-                if callable(p.pick_events_cov):
-                    picker = p.pick_events_cov
-                else:
-                    picker = p.pick_events_cov[ii]
-                events = picker(events)
-                new_count = len(events)
-                print('  Using %s/%s events for %s'
-                      % (new_count, old_count, op.basename(cov_name)))
+            picker = p.pick_events_cov  #SMB 2020-02-14
+            if type(picker) is str:
+                assert picker == 'restrict', \
+                    'Only "restrict" is a valid string for p.pick_events_cov'
+            elif picker and not callable(picker):
+                picker = picker[ii]
+            events = _read_events(p, subj, ridx, raw, picker=picker)
             # create epochs
             use_reject, use_flat = _restrict_reject_flat(reject, flat, raw)
             baseline = _get_baseline(p)
--- a/mnefun/_cov.py
+++ b/mnefun/_cov.py
@@ -143,7 +143,7 @@ def gen_covariances(p, subjects, run_indices, decim):
             raw = concatenate_raws(raws)
             this_decim = _handle_decim(decim[si], raw.info['sfreq'])
             # read in events
-            picker = p.pick_events_cov  #SMB 2020-02-14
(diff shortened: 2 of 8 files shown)


doc/overview.rst
```diff
@@ -491,6 +491,9 @@ autoreject_types : tuple
     reject trials on basis of EOG.
 reject_epochs_by_annot : bool
     If True, reject epochs by BAD annotations.
+pick_events_autoreject : callable | string | None
+    Function for picking covariance events, or the string "restrict"
```
larsoner, line 495:
```suggestion
    Function for picking autoreject events, or the string "restrict"
```

mnefun/_cov.py
```diff
@@ -143,17 +143,13 @@ def gen_covariances(p, subjects, run_indices, decim):
...
-                if callable(p.pick_events_cov):
-                    picker = p.pick_events_cov
-                else:
-                    picker = p.pick_events_cov[ii]
-                events = picker(events)
-                new_count = len(events)
-                print('  Using %s/%s events for %s'
-                      % (new_count, old_count, op.basename(cov_name)))
+            picker = p.pick_events_cov
+            if type(picker) is str:
```
larsoner, line 147: More standard / better practice generally as it allows subclasses to pass through
```suggestion
            if isinstance(picker, str):
```

mnefun/_scoring.py
```diff
@@ -104,15 +104,28 @@ def extract_expyfun_events(fname, return_offsets=False):
     return these_events, resps, orig_events
 
 
-def _read_events(p, subj, ridx, raw):
+def _pick_events(events, picker):
+    old_cnt = sum(len(e) for e in events)
+    events = picker(events)
+    new_cnt = len(events)
+    print(f'  Using {new_cnt}/{old_cnt} events.')
```
larsoner, line 111: I'm surprised the right indentation level is not at least 4 spaces, does this print properly?

mnefun/_scoring.py
```diff
@@ -104,15 +104,28 @@ def extract_expyfun_events(fname, return_offsets=False):
...
     events = list()
     for fname in get_event_fnames(p, subj, ridx):
-        these_events = read_events(fname)
-        if len(np.unique(these_events[:, 0])) != len(these_events):
-            raise RuntimeError('Non-unique event samples found in %s'
-                               % (fname,))
+        these_events = read_events(fname, include=ids)
+        # if len(np.unique(these_events[:, 0])) != len(these_events):
+        #     raise RuntimeError('Non-unique event samples found in %s'
+        #                        % (fname,))
```
larsoner, line 128: Rather than commenting this out, it seems like this should just be moved down after the `events = _pick_events(...)` conditional below

NeuroLaunch: That's fine - I'll make that change.  I just wasn't sure it is necessary at all if, just a few lines below, the concatenated events are checked for unique times.

larsoner: ohhh that happens already. Then yes this is not necessary, just delete these lines rather than commenting them

mnefun/_scoring.py
NeuroLaunch: I modified this line from previous old-style formatted string.  I think it's supposed to be slightly indented to stand apart from other logging text.  It seems to print as expected.

mnefun/_scoring.py
NeuroLaunch: Actually, I recall now that the reason I didn't keep the time check within the fname loop is in case picker IS a function and the read_events() line crashes when include=None.  Can we leave it out for now and assume the post-concatenation check is sufficient?

NeuroLaunch: Ah, sorry, I was in the middle of typing when our meeting started.  I'll delete the lines and repush.

larsoner:
@NeuroLaunch should I look again?

NeuroLaunch:
I'm going to make a one-line push to clarify an error code in ._yaml.  If you don't object to it, go ahead and merge.

larsoner:
Thanks @NeuroLaunch !

mnefun/_epoching.py
```diff
@@ -146,7 +150,8 @@ def save_epochs(p, subjects, in_names, in_numbers, analyses, out_names,
             write_hdf5(hdf5_file, use_reject, overwrite=True)
         else:
             use_reject = _handle_dict(p.reject, subj)
-        # create epochs
+        # read in events and create epochs
+        events = _read_events(p, subj, run_indices[si], raw, picker='restrict')
```
larsoner, line 154: Looking at this again, I think this is a backward compatibility break @NeuroLaunch, `_read_events` used to give back all events, no?

larsoner: Maybe it's actually a bugfix to do this actually...

NeuroLaunch: It's a bug fix at least for the way I set up my event list.  But I think it should be compatible with previous uses, as the line subsequent to 154 that creates the (final saved) epochs is limited to those in `event_id=old_dict`, which are the same ids enforced by 'restrict'.

NeuroLaunch: Also, in reviewing Ashley's soma scripts, she also has unused event ids in the stored event list.  So, implementing a restricted epoching during the cov step may give a better tmin:tmax estimate of covariance since only the intended events will be used.  (If the tmin:tmax (or is it bmin:bmax for covariance?) is particularly long, excessive events may lead to considerable redundancy in the time samples used to estimate covariance, which would lead to unwanted bias.)

PR #325, opened 2021-02-23
BUG: Handle additional entries

larsoner:
1. If there are multiple sets of files discovered, it's a problem for `fetch_raw=True`. I am hitting this with GenZ for example where some preprocessing was done in the same folder as the acquisition files on `brainstudio`, which means two sets of "original" `_raw.fif` files are found. This adds an `acq_exclude` that allows a list of substrings to look for to exclude. In principle this might also help if a subject is run twice on a given experiment -- you can exclude their unused/old acquisition by folder name (run date) for example.
2. Tweaks the `rsync` command to be NFS-friendly by not trying to set folder mtime and by not being strict about permissions set on the transferred files.

### DIFF
--- a/mnefun/_fetching.py
+++ b/mnefun/_fetching.py
@@ -47,9 +47,21 @@ def fetch_raw_files(p, subjects, run_indices):
                           % remote_fnames[:1])
         # make the name "local" to the acq dir, so that the name works
         # remotely during rsync and locally during copyfile
-        remote_dir = [fn[:fn.index(op.basename(fn))]
-                      for fn in remote_fnames][0]
-        remote_fnames = [op.basename(fname) for fname in remote_fnames]
+        remote_dirs = sorted(set(
+            fn[:fn.index(op.basename(fn))] for fn in remote_fnames))
+        # sometimes there is more than one, for example if someone has done
+        # some processing in the acquistion dir
+        orig_remotes = remote_dirs
+        remote_dirs = [remote_dir for remote_dir in remote_dirs
+                       if not any(x in remote_dir for x in p.acq_exclude)]
+        if len(remote_dirs) != 1:
+            raise IOError('Unable to determine correct remote directory, got '
+                          f'candidates:\n{remote_dirs}\n'
+                          f'pruned from\n{orig_remotes}')
+        remote_dir = remote_dirs[0]
+        del orig_remotes, remote_dirs
+        remote_fnames = sorted(set(
+            op.basename(fname) for fname in remote_fnames))
         want = set(op.basename(fname) for fname in fnames)
         got = set(op.basename(fname) for fname in remote_fnames)
         if want != got.intersection(want):
@@ -60,8 +72,8 @@ def fetch_raw_files(p, subjects, run_indices):
                           'Likely split files were found. Please confirm '
                           'results.')
         print('  Pulling %s files for %s...' % (len(remote_fnames), subj))
-        cmd = ['rsync', '-ave', 'ssh -p %s' % p.acq_port,
-               '--prune-empty-dirs', '--partial',
+        cmd = ['rsync', '-avOe', 'ssh -p %s' % p.acq_port,
+               '--no-perms', '--prune-empty-dirs', '--partial',
                '--include', '*/']
         for fname in remote_fnames:
             cmd += ['--include', op.basename(fname)]
--- a/mnefun/_fetching.py
+++ b/mnefun/_fetching.py
@@ -2,6 +2,7 @@
 
 import os
 import os.path as op
+import re
 import shutil
 import subprocess
 import warnings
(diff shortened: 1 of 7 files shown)


larsoner:
Okay this allowed all subject data to be pulled in properly. @NeuroLaunch can you look and merge if you're happy?

NeuroLaunch:
Sure, I'll set up an acquisition scenario and test it out soon.

NeuroLaunch:
OK, I tried it out on a pseudo-subject with subdirectories "209999" and "208888" containing identical raw file names.  I could exclude the files in 208888 by setting an appropriate `acq_exclude` parameter.

My main concerns is that 1) `acq_exclude` does not support a per-subject dictionary and 2) being an exclusion regex string, it can be hard to specify all directories that are _not_ the single one you're interested in.  Together this could make it tricky to implement, e.g., your scenario in which a user wants to exclude unused/old acquisition by run date.  Also, what happens if run "_01" was collected on one date and "_02" was collected on another?  In the PR code, `remote_dirs` can't have multiple listings.

Otherwise, the code passed ok.  I'll postpone the day when I finally hit that MERGE button and defer to you on next steps.  Thanks for asking me to review.

larsoner:
> 1) acq_exclude does not support a per-subject dictionary

I think regex operating on the full directory path actually helps us here because it can be (nearly) as arbitrarily precise or as general as you want. For example you can do `genz109_9a/010203` and it will work just to exclude `010203` dir for `genz109_9a` and no other subject, and you can also do `genz_[0-9]+_[0-9]+` to exclude any directory with the form `genz_<number>_<number>a`. And I wouldn't be suprised if we could turn an exclude into an include via a negative lookahead assertion or something, but my regex is not that good and it seems overkill...

>  and 2) being an exclusion regex string, it can be hard to specify all directories that are not the single one you're interested in

... in practice I'm not sure we will need this -- even for genz which was a bit of a mess for ~100 subjects I only needed to specify two regexs. FWIW we already have "multiple include" support because `params.acq_dir` is a list of str paths to search, so you could technically specify the specific directory for each subject and it would come close to being an "include list" rather than "exclude list". But using a single broad include via `acq_dir` and a handful of excludes seems to work well enough already, because you basically try to let it do its thing for all subjects, and when you hit a problem, you just add an exclusion (hopefully).

mnefun/_fetching.py
```diff
@@ -47,9 +53,23 @@ def fetch_raw_files(p, subjects, run_indices):
                           % remote_fnames[:1])
         # make the name "local" to the acq dir, so that the name works
         # remotely during rsync and locally during copyfile
-        remote_dir = [fn[:fn.index(op.basename(fn))]
-                      for fn in remote_fnames][0]
```
larsoner, line 51: > Also, what happens if run "_01" was collected on one date and "_02" was collected on another? In the PR code, remote_dirs can't have multiple listings.

See highlighted code -- I think we only ever used a single directory on `master`, too, the difference is that on `master` we (arbitrarily and silently) chose to use the first directory name rather than choosing one in a principled way. So we can figure out how to use multiple directories in another PR if necessary.

larsoner:
> Otherwise, the code passed ok. I'll postpone the day when I finally hit that MERGE button and defer to you on next steps. Thanks for asking me to review.

The "GitHub way" is that you made a review, I respond, and we iterate until you're happy then you hit the button. So I've responded, the next step is for you to decide if you're convinced by what I said, and if you are you hit the button, otherwise we can continue to discuss / I push some commits etc.

mnefun/_scoring.py
```diff
@@ -117,12 +118,18 @@ def _read_events(p, subj, ridx, raw, picker=None):
...
     events = list()
     for fname in get_event_fnames(p, subj, ridx):
-        these_events = read_events(fname, include=ids)
+        # gracefully handle empty events (e.g., resting state)
+        with open(fname, 'r') as fid:
+            content = fid.read().strip()
+        if not content:
+            these_events = np.empty((0, 3), int)
+        else:
+            these_events = read_events(fname, include=ids)
```
larsoner, line 132: Added this to handle writing an empty events list (which should be supported, for example when doing a resting-state run that will be analyzed separately)

mnefun/_scoring.py
```diff
@@ -109,6 +109,7 @@ def _pick_events(events, picker):
     events = picker(events)
     new_cnt = len(events)
     print(f'  Using {new_cnt}/{old_cnt} events.')
+    return events
```
larsoner, line 112: ... and fixed this regression from #323 :)

NeuroLaunch: Oh, good catch.  I apparently didn't thoroughly test having a proper 'picker' function.

NeuroLaunch:
>> And I wouldn't be suprised if we could turn an exclude into an include via a negative lookahead assertion or something, but my regex is not that good and it seems overkill...

I actually tried what I thought was a negative regex pattern to accept _only_  subdir 209999 (in my trivial example), but it didn't work.  IIRC the error was in the remote_fnames comprehension, where a regex match was found but was empty ('').  As you say, though, this can go in a later PR if needed.

>> params.acq_dir is a list of str paths to search, so you could technically specify the specific directory for each subject and it would come close to being an "include list" 

Ah, I didn't know that could have multiple directories.  This is great, and I agree it's a good solution for setting up includes + rare excludes.

I'll look over one more time and probably go ahead and hit \<merge\>.  Thanks!

mnefun/_scoring.py
NeuroLaunch: I double checked that an empty events handed to `concatenate_events` is acceptable, so I think we're good to go.

PR #327, opened 2021-02-25
fix: parameter order

### LINKED ISSUE #326: Wrong parameter order in call

ktavabi:
Fixes #326


larsoner:
Thanks for the fix, this must have slipped in when I made #317 !

BTW this reminds me that we should put on the TODO list to switch to using eSSS using these projectors, it should give us better SNR and less rank reduction. But we can do that once everything works :)

PR #328, opened 2021-02-25
_handle_dict call for p.ecg_t_lims

nordme:
Allows individualized subject parameters for the ecg event finding time limits.


mnefun/_ssp.py
```diff
@@ -285,7 +285,7 @@ def do_preprocessing_combined(p, subjects, run_indices):
                   % op.sep.join(bad_file.split(op.sep)[-3:]))
             bad_file = None
 
-        ecg_t_lims = p.ecg_t_lims
+        ecg_t_lims = _handle_dict(p.ecg_t_lims, subj)
```
ktavabi, line 288: These limits are supposed to be floats so `mne` doesn't balk. Are we going to convert everything to `dict` to get this `yaml` business to work? I've had to flip flop 10-20 parameters from `yaml` to explicit `mnefun.params` expressions just to get through one subject through `SSP` and now I have a 100 lines long YAML file and a 100 lines long MNEFUN script. How is this better, again? @NeuroLaunch

larsoner: If you look at what `_handle_dict` [does](https://github.com/LABSN/mnefun/blob/master/mnefun/_utils.py#L74-L81):
```
def _handle_dict(entry, subj):
    out = entry
    if isinstance(entry, dict):
        try:
            out = entry[subj]
        except KeyError:
            pass
    return out
```
Hopefully you would agree that it will not break things if `p.ecg_tlims` is a float. Basically it allows dict-ness but does not force it.

> I've had to flip flop 10-20 parameters from yaml to explicit mnefun.params expressions just to get one subject through SSP

This should not be the case, something is either wrong with the `mnefun` code or how you are translating your parameters. If you can fix the `badbaby` code to use the single, non-looped path (I would consider this a prerequisite for converting to YAML) and then open a WIP PR to convert to YAML then I can comment on where the issues are. If this is a pain in terms of GitHub gymnastics, I'm happy to "put my changes back" in one PR that makes a single flat `py` file with no loops, then do a draft to convert it to YAML conversion in another.

This latter part is basically what this PR was, the translation wasn't too bad, and the `analysis_mnefun.py` script ends up being very short:

https://github.com/LABSN/mnefun/pull/270/files#diff-7aaebcfef6318ee3bec53bec525f8b3b39ef1e2fedce4fbbc8daef3db3aed939

NeuroLaunch:
While this PR is covering a parameter update, I noticed that _p.trans_to_ written as a list isn't interpreted properly in _sss.py.  It wants that list to be a tuple, which is an easy fix in code but per Kam's comment, the initial parameter loading step should probably handle this situation.  (I tried a leading "!!python/tuple" in the .yml file, but the SafeContructor yaml loading makes it a string.). Probably the easiest change is to force a numeric _.trans_to_ list to be a tuple in _sss.py and anywhere else it's used.

ktavabi:
FTR it was `egg_t_lims` tuple in the `yaml` that triggered a `TypeError` in `mne.epochs`. So far I've had to move the following parameters out of the `yaml` file.

> params.run_names = ["%s_mmn", "%s_am", "%s_ids"]
> params.runs_empty = ["%s_erm"]
> params.subject_run_indices = [[0,1,2]]* len(params.subjects)
> params.flat = dict(grad=1e-13)
> params.auto_bad_flat = dict(grad=1e-13)
> params.ssp_ecg_reject = dict(grad=np.inf, mag=np.inf)

[These types of issues are not unheard of](https://stackoverflow.com/questions/39553008/how-to-read-a-python-tuple-using-pyyaml). I don't know what the game plan is, i.e. whether this ENH makes it easier to transition to BIDS, otherwise it's redundant. Otherwise one could just as well insist users to put all parameters in a `ipynb` cell. IMHO `mnefun`  is turning into bloat, just my 2c.

larsoner:
> it was egg_t_lims tuple in the yaml

I think this is the problem -- `yaml` has no notion of `tuple` so what you accidentally gave it was a string:
```
>>> yaml.load('x: [0, 1]')
{'x': [0, 1]}
>>> yaml.load('x: (0, 1)')
{'x': '(0, 1)'}
```
So the fix is to always use lists. Almost everywhere in mne and mnefun we treat lists and tuples interchangably and where we don't, we can fix it so that it does.

> run_names = ["%s_mmn", "%s_am", "%s_ids"]

For these I'm guessing it's a string escape problem or something. Reading about the YAML spec it should be possible to figure it out...

> params.subject_run_indices = [[0,1,2]]* len(params.subjects)

This is probably the same as omitting it entirely, or using `None` (`null` in YAML). But if you end up with some subjects that have some and some that don't, you can also use the `defaultdict` functionality (like what this PR actually implicitly implements for ecg_tlims) and do something like:
```
    subject_run_indices: {
      __default__: [0, 1, 2],
      subj_01: [0, 1],
      }
```
> params.flat = dict(grad=1e-13)

This should be fine as `flat: {grad: 1.e-13}`. The `.` is important IIRC, otherwise again it's a str -- using something with good YAML syntax highlighting might help here (at least in vscode str and numerical and null are all shown differently)

> i.e. whether this ENH makes it easier to transition to BIDS

Which ENH? I think there are two things going on:

1. This PR adds more of our standard (at least for the last couple of years?) dict-handling functionality, just for a new parameter. So I'll merge (thanks @nordme !)
2. Somehow we have gotten into a discussion on YAML that seems a bit unrelated -- we can open a separate issue to discuss this if you want. But the short version is that @drammock wanted this because it was easier to maintain and follow (?), no need to transition if you want. I'm not sure if it will help with the transition to BIDS, would have to look at mne-study-template (how I would eventually transition to BIDS) to see. But it's not the primary motivation, so if you're putting in effort to move to YAML for BIDS, I would say don't bother for now. The python-based interface is still valid to use, no need to get into the "bloat"ed YAML parts if you are unhappy with them!

NeuroLaunch:
I can put the `trans_to` fix in a separate PR, unless I was mistaken about the nature of that error (I went back to "twa" value anyway).  As for yaml adoption in general, I just transitioned Ashley from script-based to yaml-based parameter settings and it went smoothly.  There is a lot of syntax attention to maintain, but that's true anyway with Python (_cf_. my never-ending war with indentations).

larsoner:
I can't quitie remember how we handle `trans_to` offhand, but if it doesn't handle tuple and list the same it probably should, unless there is some comment in the code as to why we don't...

PR #330, opened 2021-03-02
fix-duration

ktavabi:
Quick fix for fixed length epoching in `_ssp`

### DIFF
--- a/mnefun/_ssp.py
+++ b/mnefun/_ssp.py
@@ -413,6 +413,8 @@ def do_preprocessing_combined(p, subjects, run_indices):
         events = fixed_len_events(p, raw_orig)
         reject = _handle_dict(p.reject, subj)
         use_reject, use_flat = _restrict_reject_flat(reject, flat, raw_orig)
+        if p.tmin is None or p.tmax is None:
+            p.tmin, p.tmax = (_get_baseline(p)[0], 1.)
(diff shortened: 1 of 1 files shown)


mnefun/_ssp.py
```diff
@@ -413,6 +413,8 @@ def do_preprocessing_combined(p, subjects, run_indices):
         events = fixed_len_events(p, raw_orig)
         reject = _handle_dict(p.reject, subj)
         use_reject, use_flat = _restrict_reject_flat(reject, flat, raw_orig)
+        if p.tmin is None or p.tmax is None:
+            p.tmin, p.tmax = (_get_baseline(p)[0], 1.)
```
larsoner, line 417: To mirror MNE usually tmin and tmax [should be float](https://labsn.github.io/mnefun/overview.html#epoching-epoching-parameters) and `p.baseline` can have `None`s in it. Getting tmin and tmax from `baseline` seems backward. What's the use case? I would have expected something like:
```
params.tmin = 0
params.tmax = 1
params.baseline = (None, None)  # or maybe just None
```
or so

ktavabi: Makes sense. I must have hit an error without defining an epoching window but a baseline `(-.1, None)`.
Feel free to close.

larsoner:
Okay I'll close this but if you hit a use case where defining `tmin / tmax` as float does not work properly (or `baseline` does not inherit them properly with `None`s) then let's revisit

PR #333, opened 2021-03-10
FIX: No singleton times

### LINKED ISSUE #332: TypeError: 'numpy.float64' object is not iterable
@larsoner I'm curious to see if either you or Mark hit this error when setting times='peaks' during reporting.

>   File “/home/ktavabi/Github/badbaby/badbaby/processing/run_mnefun.py”, line 161, in <module>
>     mnefun.do_processing(
>   File “/home/ktavabi/Github/mnefun/mnefun/_mnefun.py”, line 521, in do_processing
>     outs[ii] = func(p, subjects, structurals, run_indices)
>   File “/home/ktavabi/Github/mnefun/mnefun/_report.py”, line 868, in gen_html_report
>     times = _get_memo_times(
>   File “/home/ktavabi/Github/mnefun/mnefun/_report.py”, line 271, in _get_memo_times
>     memo[key] = _peak_times(evoked, cov)
>   File “/home/ktavabi/Github/mnefun/mnefun/_report.py”, line 306, in _peak_times
>     ‘,’.join(‘%0.3f’ % t for t in times),), end=‘’)
> TypeError: ‘numpy.float64’ object is not iterable

larsoner:
Closes #332

@ktavabi can you see if it fixes your problem?

### DIFF
--- a/mnefun/_report.py
+++ b/mnefun/_report.py
@@ -299,7 +299,7 @@ def _peak_times(evoked, cov, max_peaks=5):
     times = peaks[prms.argsort()[::-1]][:npeaks]
     times.sort()
     if not len(times):  # guarantee at least 1
-        times = gfp.argmax()
+        times = [gfp.argmax()]
     times = evoked.times[times]
     print('    Whitened GFP peak%s (%s):   %s\n                         '
           % (_pl(times), evoked.comment,


mnefun/_report.py
```diff
@@ -299,7 +299,7 @@ def _peak_times(evoked, cov, max_peaks=5):
     times = peaks[prms.argsort()[::-1]][:npeaks]
     times.sort()
     if not len(times):  # guarantee at least 1
-        times = gfp.argmax()
+        times = [gfp.argmax()]
```
ktavabi, line 302: Would tuple work as well?

ktavabi:
> Closes #332
> 
> @ktavabi can you see if it fixes your problem?

👍

mnefun/_report.py
larsoner: Tuples and lists are often interchangeable, but one notable time they are not is in `__getitem__` as the tuple gets parsed as a set of arguments:
```
>>> x = np.zeros(1)
>>> x
array([0.])
>>> x[0]
0.0
>>> x[(0,)]
0.0
>>> x[[0]]
array([0.])
```
So a tuple of `(0,)` gets processed the same way as `0` rather than like `[0]`.

PR #335, opened 2021-03-30
Use bmin and bmax for covariance calculation to match docs

### LINKED ISSUE #334: compute_covariance doesn't use bmin/bmax
The `baseline` parameter for Epoching states that "`params.bmin` and `params.bmax` will always be used for covariance calculation. This is useful e.g. when using a high-pass filter and no baselining is desired (but evoked covariances should still be calculated from the baseline period)", which it is currently not doing.

Will fix by changing _cov.py to use `bmin` and `bmax`

mdclarke:
fixes #334


larsoner:
Thanks @mdclarke !

PR #336, opened 2021-04-06
ENH: Add support for regex reject by annot

larsoner:
Completely untested, feel free to test @drammock ! In theory you should get a nice `print` statement when you pass a `str`.

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -502,8 +502,9 @@ autoreject_types : tuple
     Default is ('mag', 'grad', 'eeg'). Can set to ('mag', 'grad', 'eeg',
     'eog) to use EOG channel rejection criterion from autoreject module to
     reject trials on basis of EOG.
-reject_epochs_by_annot : bool
-    If True, reject epochs by BAD annotations.
+reject_epochs_by_annot : bool | str
+    If True, reject epochs by BAD annotations. If str, will reject epochs by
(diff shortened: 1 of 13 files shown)


larsoner:
@drammock do you want to try running the unified PreK pipeline on this branch before I merge? I'm tempted to merge as-is and fix whatever we need in a follow-up PR. I put in a bunch of fixes needed to work with Kam's data.

drammock:
go ahead and merge, I'm busy with other aspects of PREK right now, may be a while before we're ready to run the unified pipeline.

ktavabi:
This merged appears to have failed b/c of Travis. @larsoner I am now running your `reject` branch on badbaby data.

larsoner:
I think the code churn / increased difficulty of proper `git blame` from black is probably not worth the benefits it brings in terms of uniformity. Maybe we should do it in a separate PR for the whole repo at some point but I'd rather lag than lead MNE in adopting that convention

ktavabi:
> Maybe we should do it in a separate PR for the whole repo at some point but I'd rather lag than lead MNE in adopting that convention

SGTM But I am not sure why my local modifications got merged into your branch?! 😬

larsoner:
> SGTM But I am not sure why my local modifications got merged into your branch?! 

Must have been a `git push` done by some software you are using (codestream?)

PR #343, opened 2021-05-10
BUG: Fix handling of empty evoked

larsoner:
Closes https://github.com/ktavabi/badbaby/issues/15

@ktavabi can you give it a try? There might be other errors down the line, but I think this should fix your empty-report issue.


ktavabi (approved):
Works for me, thanks!

PR #346, opened 2021-07-09
suppress plots during _raise_bad_epochs

drammock:
helpful when running on a headless server...  currently things hang with a `qt.qpa.xcb: QXcbConnection: XCB error: 128 (Unknown), sequence: 417, resource id: 0, major code: 131 (Unknown), minor code: 2` and the `RuntimeError` with the helpful message never appears.


larsoner:
Makes sense to me, thanks @drammock !

Issue #347, opened 2021-07-21
BUG: Support EEG-only data


larsoner:
For MMF for example we need:

1. Bad channel detection for EEG
2. Fix forward modeling for EEG (@nordme)
3. Add in support for calculating covariance based on the raw data, not epoched or empty room covs (@larsoner)
4. Fix report generation (@larsoner)
5. Fix the status updates (@larsoner)

And later once TPS is finalized we can add support for processing EEG data using it (for EEG-only and M/EEG data). For now we'll leave it out...

PR #348, opened 2021-07-22
ENH: Add copy-and-skip for EEG-only data

larsoner:
WIP for EEG-only data / #347


larsoner:
@nordme @drammock ready for review/merge from my end, makes EGI-MMF EEG-only data processing work at my end!

PR #350, opened 2021-07-29
use delayed proj when epoching during cov computation

### LINKED ISSUE #349: epoching during cov computation fails because proj=False
I have a subject in the F2Fconn cohort (number 116) where epoching succeeds, but covariance computation fails, because during cov computation the Epochs constructor is called with a hardcoded `proj=False` ([here](https://github.com/LABSN/mnefun/blob/381823a9d26fdd189c4873f7ca1bed9c94f143d9/mnefun/_cov.py#L151)) and 100% of the epochs are getting rejected (267 of them survive if `proj=True`).  I have `epoching: autoreject_thresholds: True` in the config, and I think that means the values in `reject` will have been computed by `autoreject` *with the projectors active / applied* so they're really not appropriate to use when the projectors are turned off.

@NeuroLaunch @mdclarke did you run into this issue in your earlier work with this data?

@larsoner what is the logic behind using `proj=False` instead of `'delayed'` here?

drammock:
closes #349 

Waiting for confirmation from @larsoner that this is the right fix, but it's an easy enough fix that I figured I'd tee it up anyway just in case.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -12,7 +12,7 @@ Changelog
 2021
 ^^^^
 - 2021/07/29:
-    Change ``proj=True`` to ``proj='delayed'`` when epoching during covariance
+    Change ``proj=False`` to ``proj='delayed'`` when epoching during covariance
     computation, to accord with the fact that the rejection thresholds may have
     been determined from projected data.
 - 2021/05/03:
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -11,6 +11,10 @@ Changelog
 
 2021
 ^^^^
+- 2021/07/29:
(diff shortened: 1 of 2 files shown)


doc/whats_new.rst
```diff
@@ -11,6 +11,10 @@ Changelog
 
 2021
 ^^^^
+- 2021/07/29:
+    Change ``proj=True`` to ``proj='delayed'`` when epoching during covariance
```
larsoner, line 15:
```suggestion
    Change ``proj=False`` to ``proj='delayed'`` when epoching during covariance
```

larsoner:
Thanks @drammock !

PR #351, opened 2021-10-04
_concat_resamp_raws fix

nordme:
mnefun/_reports began throwing an error when trying to concatenate raw files for the purpose of plots such as the raw segments and raw psd plots. Apparently, mne's raw.info checking before concatenation has become more stringent. 

This PR adds an argument to the `_concat_resamp_raws` function called `set_dev_head_t`. Default is `False`, but if `True`, `raw.info['dev_head_t']` from the first raw is used for all subsequent raws. 

@larsoner This is working well for the data I've tested it on (the new Genz data).


larsoner:
Thanks @nordme !

Issue #355, opened 2022-02-04
ENH: Add OTP processing option


mdclarke:
Add option to process data with OTP - @larsoner I can work on this and add a pull request soon

PR #356, opened 2022-02-04
ENH: Add OTP processing option

mdclarke:
adresses #355


larsoner (commented):
@mdclarke looks like it's close! I pushed some commits to fix the CIs. CircleCI was broken and Travis wasn't running, so I fixed those. Then in terms of your changes, I had to change the order of some stuff to make the YAML match the `overview.rst`.

I think what's left is:

1. Update the [flowchart](https://407-17719360-gh.circle-artifacts.com/0/html/overview.html#flow-chart) by changing https://github.com/LABSN/mnefun/blob/main/mnefun/_flow.py
2. Make it so that the SSS step uses either the raw files (if otp_dur is None) or the OTP'ed files (if otp_dur is not None)

PR #361, opened 2023-03-29
fix except -> else when importing for mark_flat

drammock:
Today I hit
```
NameError: name 'mark_flat' is not defined
> /opt/mne/fun/mnefun/_sss.py(689)_python_autobad()
    687     del skip_stop, skip_start
    688     picks_meg = pick_types(raw.info, meg=True)
--> 689     mark_flat(raw, picks=picks_meg, verbose=False)
```

because `try: mne.preprocessing.annotate_amplitude` succeeded, but then it was never used in defining `mark_flat` because the `else` clause was wrongly coded as an `except` clause.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
       - checkout
       - run: sudo apt update && sudo apt install graphviz-dev graphviz
       - run: python -m pip install --quiet --upgrade --user pip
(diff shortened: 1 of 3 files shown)


drammock:
note to self about CI failures:

```
LookupError: setuptools-scm was unable to detect version for /tmp/pip-req-build-ni3gnmdw.
      
Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.
      
For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
```

larsoner:
I can push a quick fix

PR #362, opened 2023-03-29
select ct, cal files

nordme:
This is a fix so mnefun selects the correct calibration and cross talk files for running maxfilter in python. This code should: a) preserve the ability of the user to provide custom files either for ct or cal, b) if no custom files are provided, choose whether to use Vectorview or Triux files based on raw.info['description'] (which encodes which machine was used to record the raw file).


nordme:
@drammock So, I haven't run checks like pyflakes, etc yet, but I just went ahead and opened the PR so you can see it

larsoner:
I was thinking of selecting based on `info['meas_date']` but as long as we don't mess with `'description'` during acquisition (which I think you can in principle but we probably don't in practice) I think this is fine. We can always change to another method later if needed so +1 for this method if it works for you @nordme !

mnefun/_sss.py
```diff
@@ -396,15 +396,25 @@ def _get_origin(p, raw):
...
-    if getattr(p, 'ct_file', 'uw') == 'uw':
-        ct_file = op.join(_data_dir, 'ct_sparse.fif')
     else:
+        if raw.info['description'].startswith('TRIUX'):
+            cal_file = op.join(_data_dir, 'sss_cal_triux.dat')
+        else:
+            assert raw.info['description'].startswith('Vectorview')
+            cal_file = op.join(_data_dir, 'sss_cal.dat')
+
+    if not getattr(p, 'ct_file', 'uw') =='uw':
```
drammock, line 409:
```suggestion
    if not getattr(p, 'ct_file', 'uw') == 'uw':
```

drammock:
needs merge with main to pass CIs.  I'm getting `Permission denied` even though it looks like the "maintainers may edit" box is checked.  @nordme could you pull my flake8 fix, merge in upstream/main, then push?

drammock:
🙏🏻 @nordme!

PR #364, opened 2023-03-31
Update mnefun._reports to work with current mne-python

nordme:
I think good_hpi_count, raw_segments, and psd under mnefun._reports could all use some help. 

I'm starting with tossing the the mne.fixes call from good_hpi_count (`time_kwargs` is unused).

### DIFF
--- a/mnefun/_report.py
+++ b/mnefun/_report.py
@@ -998,7 +998,7 @@ def gen_html_report(p, subjects, structurals, run_indices=None):
                                         # old MNE had separate vmin, vmax
                                         topomap_args.update(
                                             zip(('vmin', 'vmax'),
-                                            topomap_args.pop('vlim')))
+                                                topomap_args.pop('vlim')))
                                         fig = this_evoked.plot_joint(
                                             times, **kwargs)
                                 assert isinstance(fig, plt.Figure)
--- a/mnefun/_report.py
+++ b/mnefun/_report.py
@@ -40,7 +40,6 @@ SQ2STR = '×√2'
 @contextmanager
 def report_context():
     """Create a context for making plt and mlab figures."""
-    mne.viz.set_browser_backend('matplotlib')
     import matplotlib
     import matplotlib.pyplot as plt
     style = {'axes.spines.right': 'off', 'axes.spines.top': 'off',
@@ -49,13 +48,14 @@ def report_context():
     plt.ioff()
     old_backend = matplotlib.get_backend()
     matplotlib.use('Agg', force=True)
-    try:
-        with plt.style.context(style):
-            yield
-    except Exception:
-        matplotlib.use(old_backend, force=True)
-        plt.interactive(is_interactive)
-        raise
+    with mne.viz.use_browser_backend('matplotlib'):
+        try:
+            with plt.style.context(style):
+                yield
+        except Exception:
+            matplotlib.use(old_backend, force=True)
+            plt.interactive(is_interactive)
+            raise
 
 
 # Backward compat wrappers for MNE 1.0+
--- a/mnefun/_report.py
+++ b/mnefun/_report.py
@@ -974,13 +974,22 @@ def gen_html_report(p, subjects, structurals, run_indices=None):
                                 this_evoked.add_proj(all_proj)
                             if this_evoked.nave > 0:
                                 with mne.utils.use_log_level('error'):
-                                    fig = this_evoked.plot_joint(
-                                        times, show=False, picks=picks,
-                                        ts_args=dict(proj=proj),
-                                        topomap_args=dict(
-                                            outlines='head',
-                                            vmin=min_, vmax=max_,
-                                            cmap=cmap, proj=proj))
+                                    try:
+                                        fig = this_evoked.plot_joint(
+                                            times, show=False, picks=picks,
+                                            ts_args=dict(proj=proj),
+                                            topomap_args=dict(
+                                                outlines='head',
+                                                vlim=(min_, max_),
+                                                cmap=cmap, proj=proj))
+                                    except TypeError:
+                                        fig = this_evoked.plot_joint(
+                                            times, show=False, picks=picks,
+                                            ts_args=dict(proj=proj),
+                                            topomap_args=dict(
+                                                outlines='head',
+                                                vmin=min_, vmax=max_,
+                                                cmap=cmap, proj=proj))
                                 assert isinstance(fig, plt.Figure)
                                 fig.axes[0].set(ylim=(-max_, max_))
                                 t = fig.axes[-1].texts[0]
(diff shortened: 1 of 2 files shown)


nordme:
Ok, reports are working for me now except for raw_segments. On my machine, raw.plot() gives me a `MNEQtBrowser` instance, but mnefun/_reports seems to expect a matplotlib instance. I'm turning off this section for running reports for today, but let's chat about fixing it.

drammock:
> Ok, reports are working for me now except for raw_segments. On my machine, raw.plot() gives me a `MNEQtBrowser` instance, but mnefun/_reports seems to expect a matplotlib instance.

could use a context handler `with mne.viz.use_browser_backend('matplotlib'):` or could set it for the session duration with `mne.viz.set_browser_backend('matplotlib')`.  For changes within mnefun code I'd guess the context handler is better?

nordme:
One more issue. It looks like SVG as an image format option isn't working for me for some reason. Here's what my drop log plot looks like. lol!!
 
Screenshot from 2023-03-31 16-16-46

nordme:
@drammock  Strangely, it looks like `_report_raw_segments` is run under a context handler already that should be using matplotlib. I haven't seen an obvious reason why it isn't working for me yet.

nordme:
Ok, I figured out the raw_segments issue. You can't pass a backend choice to raw.plot() directly; you can only set it in your mne-python.json configuration file on your own machine. `MNE_BROWSER_BACKEND` wasn't set in my config file, and the automatic backend selection function for mne returns qt because it checks if qt works before checking if matplotlib works.

mnefun/_report.py
```diff
@@ -40,6 +40,7 @@
 @contextmanager
 def report_context():
     """Create a context for making plt and mlab figures."""
+    mne.viz.set_browser_backend('matplotlib')
```
larsoner, line 43: you should use the https://mne.tools/dev/generated/mne.viz.use_browser_backend.html context manager so that it is only set temporarily

nordme: So, I tried use_browser_backend first, but that failed for me (I still got a qt instance). :) I didn't dig in to why it wasn't working -- thoughts?

drammock: Did you do it in a "with" clause? It won't work if it's not done that way. Even though it's within another context manager, create a second context within it for the browser backend.

nordme: Specifying my above comment: I think I had trouble because I don't have experience nesting context managers inside of other context managers. I can see it's possible because in the `report_context()` function, you do this: 

```        
try:
    with plt.style.context(style):
        yield
except Exception:
    matplotlib.use(old_backend, force=True)
    plt.interactive(is_interactive)
    raise```

to use the plt style context. But what's the functional and proper way to do that embedding with `mne.viz.use_browser_backend` ?  Then I was having doubts about how many levels of context nesting make sense! But I do see the advisability of only setting the backend temporarily.

Comment (larsoner) [reply mnefun/_report.py:43]: You can do as many levels you want
```
with a:
    with b:
        ...
```
Or if you will want them both active the whole `.. ` of code (nothing with just `a`) then you can instead do
```
with a, b:
    ...
```

Comment (nordme) [reply mnefun/_report.py:43]: Here's one example... 
Screenshot from 2023-04-01 21-18-18

No plot pops up at all. ??? Am I missing something obvious? lol

Comment (nordme) [reply mnefun/_report.py:43]: @drammock I tried this also (which, I think, is what you described)

Screenshot from 2023-04-01 21-35-45

..which gives me the following, without any plot at all. 

Screenshot from 2023-04-01 21-36-07

Comment (nordme) [reply mnefun/_report.py:43]: P.S. This is not urgent, fyi. :smile:  (I just needed to get reports working *enough* that I could run them on the subjects reprocessed with the new calibration files, and that got done Friday night.) Sorry to talk at you on the weekend -- please ignore till convenient!

Comment (larsoner) [reply mnefun/_report.py:43]: Without looking at the code... I think in report_context we probably set the backend to Agg, which is a non-GUI backend. So nothing popping up is the correct/desired behavior here!

Comment (nordme) [reply mnefun/_report.py:43]: Alrighty, friends, I propose the following fix. Basically, I think my issue is Pycharm, which tries to do some backend handling stuff when importing matplotlib. If I call `with mne.viz.use_backend_browser('matplotlib'):`  *after* importing matplotlib and forcing it to use `Agg` in the `report_context()` function, then everything works.

Comment (nordme) [reply mnefun/_report.py]: @larsoner (Sorry, I didn't see your comment earlier). `Agg` shouldn't cause your console to freeze such that you have to force quit it though, right??  :laughing:  I ran the above code from interactive python in a terminal earlier today, and all of the above code options worked perfectly from the terminal, so that makes me feel pretty confident that Pycharm is what's causing my problems!!

Comment (drammock) [reply mnefun/_report.py]: 🌩️#️⃣⁉️pycharm💩

Comment (nordme) [reply mnefun/_report.py]: @drammock  You know, you were the one who told me back in the day that checking your code in the terminal is a good troubleshooting step. Thanks for the good advice!!! :laughing:

**Comment by nordme** [comment]
So, I propose that we just use the png format for all our mnefun plots. It's immediately functional, and we can work on the mne.reports functionality separately.

**Comment by nordme** [comment]
Ok, reports are running smoothly both with my genz2 environment and with a fresh dev environment running mne 1.4, so we're in business on my end!

**Comment by larsoner** [comment]
Sounds good except that we should use webp instead of png nowadays. It produces smaller files even at lossless quality

**Comment by nordme** [comment]
Mmm, ok. How much backward compatibility do we have with webp?

**Comment by nordme** [comment]
Ok, mne.Report does its own checking to see if it can use webp, so my thought was to just use whatever it comes up with when it instantiates to control all the rest of the report building beneath that.

**Review by drammock — APPROVED** [review]
circle is complaining about missing config / oauth change. @larsoner do you know what that's about?

File: mnefun/_report.py [hunk 2]
```diff
@@ -263,12 +266,21 @@ def _report_raw_segments(report, raw, lowpass=None):
...
     fig.subplots_adjust(0.025, 0.0, 1, 1, 0, 0)
-    _add_figs_to_section(report, fig, section, section, image_format='png')
+    _add_figs_to_section(report, fig, section, section, image_format=img_form)
     print('%5.1f sec' % ((time.time() - t0),))
 
 
+def _gen_psd_plot(raw, fmax, n_fft, ax):
+    if hasattr(mne.time_frequency.Spectrum, 'plot'):
+        plot = raw.compute_psd(fmax=fmax, n_fft=n_fft).plot(show=False,
+                                                            axes=ax)
```
drammock, line 276: this won't work actually: if there is no `Spectrum` class at all in the MNE version, then you get `AttributeError, module 'mne.time_frequency' has no attribute 'Spectrum'`

```suggestion
    if hasattr(raw, 'compute_psd'):
        plot = raw.compute_psd(fmax=fmax, n_fft=n_fft).plot(show=False,
                                                            axes=ax)
```

